PR #5, opened 2023-07-18
add regex constraint to percent_type

joseph-sentry:
The current code that is missing the regex accepts any string for this field, but we only want strings that represent percentages or decimal numbers.


joseph-sentry:
> @joseph-sentry is this PR still live? it looks like it might have been merged already

@matt-codecov I think this still needs to get merged, but I haven't seen the issue pop up recently

matt-codecov (approved):
i wonder why i thought it had been merged. i must have been viewing the files from your branch instead of main

PR #13, opened 2023-07-26
perf: quit eagerly populating fork information in api queries

matt-codecov:
`list_repos()` returns a list of repos, and if a repo is a fork, it will have a `fork` key pointing to its source. the torngit github implementation of `list_repos()` will perform an extra API request for each fork to fill out the source's information, and these requests aren't batched/parallelized so if an org has many forks it could substantially bog down syncing. the other implementations don't do an extra network request but do populate the `fork` key

`sync_repos` only does a couple things with that `fork` key:
- upsert the forked repo and its owner to the db
- if the forked repo is private, add its ID to the list of repos the current user can see

if the forked repo is visible to the current user, it'll be in the list that `list_repos()` returns and will be handled in `sync_repos` anyway. if the forked repo is *not* visible to the current user, then it's incorrect to add it to the list of repos the current user can see.

no relationship between a repo and the repo it was forked from is saved in the db here. `list_repos()` is not used anywhere else. this PR also removes the `fork` key from the return format of torngit's `get_repository()`, but i can't find any usages of that function in the first place

(also i added vim swapfiles to gitignore)

### Legal Boilerplate

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -141,7 +141,8 @@ dmypy.json
 # Cython debug symbols
 cython_debug/
 
-# vscode settings
+# Editor junk
 .vscode/settings.json
+.*.sw?
 
 /target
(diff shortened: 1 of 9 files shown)


giovanni-guidini (approved):
LGTM

PR #16, opened 2023-07-28
fix: Consistent return types for list_repos and list_repos_using_installation

scott-codecov:
`list_repos_using_installation` was passing data back straight from GitHub.  This change mimics the format returned from `list_repos`.  Note that this is a breaking change for the `worker` - we'll need to make a few updates there when we incorporate this version of `shared`.

### DIFF
--- a/shared/torngit/github.py
+++ b/shared/torngit/github.py
@@ -350,7 +350,7 @@ class Github(TorngitBaseAdapter):
             ),
         )
 
-    async def list_repos_using_installation(self, username):
+    async def list_repos_using_installation(self, username=None):
         """
         returns list of repositories included in this integration
         """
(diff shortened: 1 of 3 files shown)


shared/torngit/github.py
```diff
@@ -350,7 +350,7 @@ async def get_repository(self, token=None):
             ),
         )
 
-    async def list_repos_using_installation(self, username):
+    async def list_repos_using_installation(self, username=None):
```
scott-codecov, line 353: `username` is not used so I made this argument optional.  Maybe we should just remove it entirely and make the corresponding change in `worker` as well.

shared/torngit/github.py
```diff
@@ -370,12 +370,27 @@ async def list_repos_using_installation(self, username):
...
+                                branch=repo["default_branch"] or "master",
+                            ),
+                        )
+                    )
+
                 if len(res["repositories"]) < 100:
                     break
 
-            for repo in repos:
-                repo["id"] = str(repo["id"])
```
scott-codecov, line 378: `worker` is currently using this but we can check `repo["repo"]["service_id"]` instead when incorporating this

matt-codecov (approved):
looks good, thank you for tidying this up and adding a test!

PR #25, opened 2023-08-15
Add minio port in install config schema

joseph-sentry:
Adds port field to minio object in install yaml schema, since its being used in `services/storage.py` in tha api

### Legal Boilerplate


matt-codecov (approved):
did this get rid of the warning log you were seeing?

joseph-sentry:
> did this get rid of the warning log you were seeing?

Yes

PR #27, opened 2023-08-16
feat: github app refresh tokens

giovanni-guidini:
Adding support for github app refresh tokens.
This is currently an opt-in feature for github apps (what we use to auth users).

closes codecov/engineering-team#163

### Legal Boilerplate

### DIFF
--- a/shared/encryption/token.py
+++ b/shared/encryption/token.py
@@ -20,10 +20,9 @@ def decode_token(_oauth: str) -> OauthConsumerToken:
     This function decrypts a oauth_token into its different parts.
     At the moment it does different things depending on the provider.
 
-    - github
-        Only stores the "key" as the entire token
     - bitbucket
         Encodes the token as f"{key}:{secret}"
+    - github
     - gitlab
         Encodes the token as f"{key}: :{refresh_token}"
         (notice the space where {secret} should go to avoid having '::', used by decode function)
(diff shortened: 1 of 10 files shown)


scott-codecov (commented):
This looks good to me.  Like we chatted about earlier I'm just a little curious about the sequence of releasing this.  Current tokens never expire and don't have a refresh token.  What's the smoothest way to roll this out?

If the currently issued tokens still never expire (and only new logins have expiring tokens + refresh tokens) then it might make sense to not raise `TorngitCantRefreshTokenError` when missing a `refresh_token`.  Might be worth testing out the transition from non-expiring token GitHub app -> expiring token GitHub app and see what happens w/ old tokens.

giovanni-guidini:
I agree that we need to test the transition.

About 
>If the currently issued tokens still never expire (and only new logins have expiring tokens + refresh tokens) then it might make sense to not raise TorngitCantRefreshTokenError when missing a refresh_token.

Well we only raise that if we try to refresh the token. If it never expires it shouldn't 401 and therefore we won't ever try to refresh it :E

matt-codecov (approved):
seconding scott's Q about rolling this out. do we want to phase out the non-expiring tokens? did we do anything like that for GitLab? looks like you addressed scott's feedback about raising `TorngitCantRefreshTokenError`, but there remain a couple dead references to it in the PR

if multiple api requests/worker tasks (particularly the latter) are spinning up at once, i'm worried they'll get stuck in a cycle of invalidating each others' tokens which will spawn a bunch of unnecessary API requests and hurt performance. i think that's worth solving, but maybe in a separate PR as it would involve changing the GitLab implementation as well. this PR is equivalent to the GitLab implementation and the tests are thorough so i'll approve

incidentally, i noticed that `_on_token_refresh()`/`get_token_refresh_callback()` is only populated in `worker/services/repository.py` and `worker/services/repo_providers.py` and is missing from `worker/services/owner.py` (and maybe other places). didn't check API, but that's a bug in worker at least

PR #28, opened 2023-08-28
chore: update gh refresh heuristic

giovanni-guidini:
Yes I know, this should have been part of codecov.shared#27

I decided to come back and restrict the requests we will try to refresh a little.
By checking the callable first we avoid app-to-server requests, which should
represent a considerable portion of requests to github, in fact.

So here's a rough list of requests that wi will NOT try to refresh:
* check if user is a student
* any request made by the codecov github app
* any request made by a repo bot
* requests for repo information made in `sync_repos` or `sync_teams`

hopefully by restricting the attempts at refreshing info we avoid unecessary reqeusts to gh.

context codecov/engineering-team#162

### Legal Boilerplate

### DIFF
--- a/shared/torngit/github.py
+++ b/shared/torngit/github.py
@@ -178,10 +178,18 @@ class Github(TorngitBaseAdapter):
                 raise TorngitServerUnreachableError(
                     "GitHub was not able to be reached."
                 )
+            # Github doesn't have any specific message for trying to use an expired token
+            # on top of that they return 404 for certain endpoints (not 401).
+            # So this is the little heuristics that we follow to decide on refreshing a token
             if (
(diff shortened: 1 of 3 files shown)


giovanni-guidini:
One more detail I'd like to add here is that GitHub doesn't send `refresh_token` in the logins unless you opt-in to that feature.
So if we don't change the current app running in production we will never try to actually refresh any tokens, (because the code will return in https://github.com/codecov/shared/blob/bbdf61a36ed253c8237f3aeec76b9bcf09cdd6a8/shared/torngit/github.py#L283)

This is also interesting because we would get a log message that can roughly be used a metric of requests that we would try to refresh the token

scott-codecov (approved):

PR #33, opened 2023-09-05
create a generic implementation for easier future changes

dana-yaish:
This PR is to have a generic implementation for analytic tools. Currently we have Segment used in worker and api, but we want to change that to another tracking tool. This PR builds a generic implementation for any tracking tool.
This PR:
- builds a manager class that has all tracking methods but leaves the actual tracking to each individual analytic tool under `track_event`. Having a manager, because at some point in time - in the near future-, we'll have to use both, the old and the new tracking tools, to debug and to make sure everything's working as expected. 
- After discussing it with Paul, we only need 5 events to track. Link is attched to this [notion doc ](https://www.notion.so/sentry/Codecov-Segment-deprecation-handling-events-62cb4c0e1d2f41cf865140d676aa02f6)
- The idea is to abstract all analytic tool details from api and worker. So in either one, we'll import `analytics_tool` from `shared.analytics_tracking` and then use `analytics_tool.whatever_tracking_method_we_have` 

many PRs will follow in worker and API.

### Legal Boilerplate

### DIFF
--- a/shared/analytics_tracking/__init__.py
+++ b/shared/analytics_tracking/__init__.py
@@ -4,17 +4,16 @@ from typing import List
 from shared.analytics_tracking.base import BaseAnalyticsTool
 from shared.analytics_tracking.manager import AnalyticsToolManager
 from shared.analytics_tracking.noop import NoopTool
+from shared.analytics_tracking.pubsub import PubSub
 from shared.analytics_tracking.segment import Segment
 
 log = logging.getLogger("__name__")
 
-__all__ = "analytics_tool"
+__all__ = "analytics_manager"
 
 
 def get_list_of_analytic_tools() -> List[BaseAnalyticsTool]:
-    return [
-        Segment(),
-    ]
+    return [Segment(), PubSub()]
 
 
 def get_tools_manager():
@@ -30,4 +29,4 @@ def get_tools_manager():
     return tool_manager
 
 
-analytics_tool = get_tools_manager()
+analytics_manager = get_tools_manager()
--- a/shared/analytics_tracking/__init__.py
+++ b/shared/analytics_tracking/__init__.py
@@ -1,171 +1,32 @@
 import logging
+from typing import List
 
-import analytics
-
-from shared.config import get_config
+from shared.analytics_tracking.base import BaseAnalyticsTool
+from shared.analytics_tracking.manager import AnalyticsToolManager
+from shared.analytics_tracking.noop import NoopTool
+from shared.analytics_tracking.segment import Segment
 
 log = logging.getLogger("__name__")
 
-BLANK_SEGMENT_USER_ID = -1
-
-# Set up
-def on_error(error):
-    log.error("Error tracking with segment", extra=dict(error=error))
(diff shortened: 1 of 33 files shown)


shared/analytics_tracking/__init__.py
```diff
@@ -1,171 +1,24 @@
...
 
-def track_manual_critical_file_labelling_added_in_YAML(repoid, ownerid, is_enterprise):
-    track_event(
-        user_id=BLANK_SEGMENT_USER_ID,
-        event_name="Impact Analysis Manual Critical File Labelling in YAML",
-        event_data={"repo_id": repoid, "repo_owner_id": ownerid},
-        is_enterprise=is_enterprise,
-    )
+def get_analytics_tool() -> BaseAnalyticsTool:
+    for tool in get_list_of_analytic_tools():
```
giovanni-guidini, line 19: I think this should be configurable in the YAML. Similar to how the [CLI loads plugins](https://github.com/codecov/codecov-cli/blob/master/codecov_cli/plugins/__init__.py#L31)
 
That will allow users to use their own tracking tool without having to change our code much. (Although it can be argued that if you are hosting for yourself you don't _need_ a tracking tool, but whatever).

Then of course we will have our own implementations like you have there and use them as default

shared/analytics_tracking/__init__.py
```diff
@@ -1,171 +1,24 @@
...
-
-
-def track_show_critical_paths_added_in_YAML(repoid, ownerid, is_enterprise):
-    track_event(
-        user_id=BLANK_SEGMENT_USER_ID,
-        event_name="Impact Analysis Show Critical Files in YAML",
-        event_data={"repo_id": repoid, "repo_owner_id": ownerid},
-        is_enterprise=is_enterprise,
-    )
+__all__ = "analytics"
```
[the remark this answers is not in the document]
dana-yaish: This controls what to be exported from the analytics_tracking module when using * in an import statement. It doesn't make a huge difference, but it's a way to say that the only thing you need from this folder is the analytics variable

shared/analytics_tracking/noop.py
```diff
@@ -0,0 +1,14 @@
...
+log = logging.getLogger("__name__")
+
+
+class NoopTool(BaseAnalyticsTool):
+    @classmethod
+    def is_enabled(cls):
+        return False
+
+    def track_event(self, user_id, event_name, *, is_enterprise, event_data={}):
+        log.warning("Analytics tool is not enabled. Please check your configuration.")
```
giovanni-guidini, line 14: I think this warning should be moved to `get_analytics_tool`. Otherwise it might spam users that opt to not have the analytics set (as we discussed, they might not care / not need it).

Then this method can do nothing

giovanni-guidini (commented):
LGTM. Just one final comment that I'd like to get your opinion on.

shared/analytics_tracking/noop.py
dana-yaish: yup you're right

giovanni-guidini (approved):

PR #37, opened 2023-09-12
introduce pubsub, add tests, refactor code, remove unneeded code in m…

dana-yaish:
…anager

### Legal Boilerplate

### DIFF
--- a/shared/analytics_tracking/base.py
+++ b/shared/analytics_tracking/base.py
@@ -1,5 +1,7 @@
 from abc import ABC, abstractmethod
 
+from shared.analytics_tracking.events import Event
+
 
 class BaseAnalyticsTool(ABC):
     BLANK_USER_ID = -1
@@ -9,8 +11,5 @@ class BaseAnalyticsTool(ABC):
     def is_enabled(cls):
         raise NotImplementedError()
 
-    def track_event(self, user_id, event_name, *, is_enterprise, event_data={}):
-        raise NotImplementedError()
-
-    def track_user(self, user_id, user_data={}, is_enterprise=False):
+    def track_event(self, event: Event, *, is_enterprise, context: None):
         raise NotImplementedError()
--- a/shared/analytics_tracking/events.py
+++ b/shared/analytics_tracking/events.py
@@ -1,4 +1,15 @@
+import logging
+from base64 import b64encode
+from datetime import datetime
 from enum import Enum
+from typing import Any, Mapping
+from uuid import uuid1
+
+import pytz
+
+from shared.helpers.date import to_timestamp
+
+log = logging.getLogger("__name__")
 
 
 class Events(Enum):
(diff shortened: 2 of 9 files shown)


shared/analytics_tracking/manager.py
```diff
@@ -15,92 +12,8 @@ def add_tool(self, tracking_tool: BaseAnalyticsTool):
...
-            if tool.is_enabled():
-                tool.track_event(
-                    user_id,
-                    event_name,
-                    is_enterprise=is_enterprise,
-                    event_data=event_data,
-                )
-
-    def track_user(self, user_id, user_data={}, is_enterprise=False):
+    def track_event(self, event_name, *, is_enterprise, event_data={}, context=None):
```
giovanni-guidini, line 15: At this point I've had Scott mentioning to me enough times that `event_data={}` has a bug in it. The reference to the dict passed is always the same (for all function calls).

The better way of doing it is:
```python
event_data: Optional[dict] = None
if event_data is None:
    event_data = {}
```

shared/analytics_tracking/events.py
```diff
@@ -7,3 +18,24 @@ class Events(Enum):
     ACCOUNT_UPLOADED_COVERAGE_REPORT = "Account Uploaded Coverage Report"
     USER_SIGNED_IN = "User Signed In"
     USER_SIGNED_UP = "User Signed Up"
+
+
+class Event:
+    def __init__(self, name: str, dt: datetime = None, **data: Any) -> None:
```
giovanni-guidini, line 24: Shouldn't `name` be of type `Events`? (which might be called `EventName` possibly)

So that I can pass for example `name=Events.USER_SIGNED_IN` directly?

shared/analytics_tracking/base.py
```diff
@@ -9,8 +11,5 @@ class BaseAnalyticsTool(ABC):
     def is_enabled(cls):
         raise NotImplementedError()
 
-    def track_event(self, user_id, event_name, *, is_enterprise, event_data={}):
-        raise NotImplementedError()
-
-    def track_user(self, user_id, user_data={}, is_enterprise=False):
+    def track_event(self, event: Event, *, is_enterprise, context: None):
```
giovanni-guidini, line 14: (as an end user) Do I need to create the event myself before calling the `track_event`?

i.e. can I do `analytics_tool.track_event(Events.USER_SIGNED_UP, event_data={...}, is_enterprise=False)` ?

dana-yaish: The idea here is that the end user will have to deal with the analytics manager, and not know anything about events. The manager will take the data and wrap it in an event object and send it to the analytic tools individually. 
We probably need to rename analytics_tool to analytics_manager because that's what it really is

shared/analytics_tracking/events.py
dana-yaish: good point. The reason for this is that I don't want to let the manager worry about valid event names whatsoever, it only passes the string that the client code sends. Then the name gets validated in the Event class

shared/analytics_tracking/manager.py
dana-yaish: didn't know that, thank you!

shared/analytics_tracking/base.py
giovanni-guidini: Good, good. I just wanted to make sure the user didn't have to know about both the manager and the events. Thanks.

shared/analytics_tracking/events.py
giovanni-guidini: I would argue that using an Enum still counts as "the manager doesn't have to worry about event names". It's the user - calling `track_event` - that needs to worry about having a valid event name and what that event name should be :E 

That being said I think your reasoning is valid and if you want to keep as is it's fine too.

giovanni-guidini (approved):

PR #41, opened 2023-09-20
feat: create Feature class for handling feature rollouts

matt-codecov:
this came up on slack yesterday and i thought it'd be fun to throw something together. let me know if you have any feedback or if we want to use a Real Product for this instead

notes:
- it can nominally support A/B tests, but our data isn't prepared to support analysis. mostly meant for simple rollouts at the moment
- increasing a rollout requires a code change and new deploy which is not ideal. an alternative could be to read the rollout proportions from redis and cache for, say, 4h,  if we have a good way to update values in production redis
- i used the 128-bit hash because it was there, but at our volumes we are probably safe from the birthday problem with 64 bits? i didn't crunch the numbers though, just going on vibes
- the bucketing logic uses floats with int truncation which a serious statistician may find introduces biases but it should hold up for now

```
# Simple feature rolled out to 20% of repos
# Override codecov repos into the feature early
# By default, features have a single variant:
#   {"enabled": FeatureVariant(True, 1.0)}
MY_FEATURE_BY_REPO = Feature(
    "my_feature",
    0.2,
    overrides={
        "{shared repo id}": "enabled",
        "{worker repo id}": "enabled",
        "{codecov-api repo id}": "enabled",
        "{gazebo repo id}": "enabled",
        "{codecov-cli repo id}": "enabled",
)
if MY_FEATURE_BY_REPO.check_value(repo_id, default=False):
    new_behavior()

# Simple A/B test rolled out to 10% of users (5% control, 5% test)
MY_EXPERIMENT_BY_USER = Feature(
    "my_experiment",
    0.1,
    variants={
        "test": FeatureVariant(True, 0.5),
        "control": FeatureVariant(False, 0.5),
    },
)
if MY_EXPERIMENT_BY_USER.check_value(user_id, default=False):
    new_behavior()
```

### Legal Boilerplate

### DIFF
--- a/shared/rollouts/features.py
+++ b/shared/rollouts/features.py
@@ -1,3 +1,14 @@
-PARALLEL_UPLOAD_PROCESSING_BY_ORG = Feature(
-    "parllel_upload_processing", 0.2, overrides={"codecov's id": True}
-)
+# This file "defines" features which actually should live in the worker. It
+# serves as a model for how this feature flag utility is recommended to be used
+# in other repositories: with a central file in which all experiments are
+# declared so that it's easier to see how much runtime variability there is in
+# the way our service behaves. Unreproducible bug report from a customer? Check
+# if there have been any recent feature changes that sound related.
+#
+# PARALLEL_UPLOAD_PROCESSING_BY_ORG = Feature(
+#     "parllel_upload_processing", 0.2, overrides={"codecov's id": True}
+# )
+#
+# LIST_REPOS_GENERATOR_BY_ORG = Feature(
+#     "list_repos_generator", 0.5, overrides={"codecov's id": True)
+# )
--- a/shared/rollouts/features.py
+++ b/shared/rollouts/features.py
@@ -1,0 +1,3 @@
+PARALLEL_UPLOAD_PROCESSING_BY_ORG = Feature(
+    "parllel_upload_processing", 0.2, overrides={"codecov's id": True}
+)
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ setup(
         "oauthlib",
         "redis",
         "typing",
+        "mmh3",
         "typing_extensions",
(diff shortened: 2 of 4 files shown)


giovanni-guidini:
This is an interesting project, thanks for putting it together.

I'm particularly interested in the `overrides` right now. Will it be possible to pass an org_id and have the feature be enabled to all of the org? (for the case we want to test things in codecov first)

Also, I don't have to define my Features and Experiments in `shared`, do I? That can be done in api or worker directly?

giovanni-guidini (approved):
I like in-house solutions 🤷‍♂️ hehehe
Thanks

(this is only my opinion, of course. We should wait to hear from others)

shared/rollouts/features.py
```diff
@@ -0,0 +1,3 @@
+PARALLEL_UPLOAD_PROCESSING_BY_ORG = Feature(
```
scott-codecov, line 1: What's this?

scott-codecov (approved):
This is nice @matt-codecov!

There's currently some %-based rollouts that are happening via the code here: https://github.com/codecov/shared/blob/071b899ae80c7a465b4e6d7d8387d9dff241e842/shared/yaml/user_yaml.py#L142C1-L142C1 (and configured here: https://github.com/codecov/k8s-v2/blob/3cc3da33deba0eb0dbd6586a11d5f034b9839acb/production.yaml#L391).  They have slightly different percentages in different envs.  If we need to continue supporting that sort of thing then maybe we come up with a convention for specifying variant proportions in the config.

shared/rollouts/features.py
matt-codecov: i wanted a "real" example feature to serve as a reference and placeholder, but for that actual experiment i'll probably put the feature definition in `worker`. i'll comment it out / explain in that file

PR #45, opened 2023-09-21
chore: Increase notify default timeout

giovanni-guidini:
These changes increase the notify default timeout to 120s instead of 60s.
I would argue that it's still a sane amount of time for notifications, and would better support customers
that use github checks heavily.

This is proposed after the investigation in codecov/internal-issues#66

EDIT
I did check with infra team that the autoscaling should handle notification queue increase - if it were to happen. I don't think there are other immediate implications...

### Legal Boilerplate

### DIFF
--- a/shared/celery_config.py
+++ b/shared/celery_config.py
@@ -151,7 +151,7 @@ class BaseCeleryConfig(object):
 
     notify_soft_time_limit = int(
         get_config(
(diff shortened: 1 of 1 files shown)


scott-codecov (approved):
Yeah I guess this seems reasonable to me.  Not entirely sure what else to do other than potentially fan out batches of checks into separate tasks.  But then you're just flooding the queue with a bunch of tiny tasks that still potentially block other notifications from happening until they're done.

PR #46, opened 2023-09-22
perf: create generator versions of `list_repos()` and optimize github's

matt-codecov:
### reviewer tips

the [first commit](https://github.com/codecov/shared/pull/46/commits/ed4b0cd6b5a7ea9e7fdc9a85998b05f65481d075) adds nothing, it just factors the logic that fetches / processes each page of repos out into separate functions. the [second commit](https://github.com/codecov/shared/pull/46/commits/6417ce189ae5982e8f36b6c299e36f6a604536c4) reuses those functions to write `_generator()` versions of the original `list_repos()` functions:
- the bitbucket and bitbucket server ones are totally identical to the originals except with `yield` instead of `return`
- the gitlab one literally calls the original and then splits the result into pages to yield
- the github ones are more complicated:
  - we pre-fetch the total number of repos so that we can fetch all the pages concurrently and yield them as they finish
  - thanks to the refactor in the first commit, `list_repos_generator()` can handle both `using_integration=True` and `using_integration=False`. `list_repos_using_installation_generator()` thus delegates to `list_repos_generator()`.

i'll use https://github.com/codecov/shared/pull/41 in worker to allow us to override specific repos in and then roll out gradually to the general population. after the rollout completes, `list_repos()` should be deleted and `list_repos_generator()` should be renamed `list_repos()`.

### details

`list_repos()` today will fetch all the visible repos, do some light munging, and then return. a new user who is syncing repos for the first time will have to wait for potentially a long while if their org has a lot of repos before they see anything at all in the UI. by yielding a page at a time, the worker's `sync_repos` task will interleave database writes and API fetches so the UI can observe incremental progress if it refreshes its viewport or whatever. this will make the UX more responsive for new customers with a lot of repos

additionally, github's graphql API lets you query the total count of visible repositories. from [my testing](https://gist.github.com/matt-codecov/6c6b9a3e652ad41103d2557761eafac0) the size it returns matches the size of the list you get back from the REST API. if we know the total number of repositories, we can pre-compute how many pages we need to fetch and then begin all the requests concurrently. so in addition to being more responsive, github syncs should complete faster overall as well.

replaces old PR https://github.com/codecov/shared/pull/24

### Legal Boilerplate


shared/torngit/gitlab.py
```diff
@@ -440,6 +440,19 @@ async def list_repos(self, username=None, token=None):
         data = [i for n, i in enumerate(data) if i not in data[n + 1 :]]
         return data
 
+    async def list_repos_generator(self, username=None, token=None):
+        """
+        Unlike GitHub and Bitbucket, GitLab has to pull a complete list of repos
```
giovanni-guidini, line 445: Don't you think that if we keep some `set()` of yielded repos we can de-dup per group and behave more like the other implementations? It would also let us be more efficient, given that we can process a page while fetching the next group.

I don't think realistically that an extra set with repo ids will use that much memory. In fact it might even save some, given that we don't have to keep the data of all repos until the end.

giovanni-guidini (commented):
Looking good. I like this in general. Question about gitlab.

shared/torngit/gitlab.py
matt-codecov: @giovanni-guidini i quickly tried that earlier but i didn't get tests passing, didn't investigate why

i think the population of onboarding gitlab users who would see the specific unimpressive (but not bugged) experience this change improves is quite small, and ultimately everybody (including them) is probably better served by the rest of the stuff on my plate. can always come back to this if we see signs that this is a common pain point

giovanni-guidini (approved):

PR #47, opened 2023-09-22
fix: workaround for ats failures

giovanni-guidini:
There's some strange env config issue around the google packages that I couldn't
quite figure out or fix.

Through many envs being built and destroyed I made this workaround locally work.
Hopefully it works in CI as well.

context (and my adventures investigating this): codecov/engineering-team#536

### Legal Boilerplate


scott-codecov (approved):
Weird - I don't really understand why this works but seems better than ATS failures

PR #51, opened 2023-09-27
fix: attempt to parse env vars into python types in config

matt-codecov:
https://github.com/codecov/internal-issues/issues/99

a value may be `True` if parsed from a yaml file but `"true"` if parsed from an env var. who knows what bugs that may cause

this PR attempts to recognize env var values that could be converted to Python types:
- bool
- int
- float

### Legal Boilerplate


shared/config/__init__.py
```diff
@@ -89,6 +90,22 @@ def load_env_var(self):
                 current[multiple_level_vars[-1].lower()] = data
         return val
 
+    def _env_var_value_cast(self, data):
+        if isinstance(data, str):
+            if data == "true":
```
JerrySentry, line 95: Thoughts on doing `data.lower()` for true/false?

Also `isdigit` handles positive ints only.

matt-codecov: > Thoughts on doing data.lower() for true/false?
it looks like pyyaml supports more than just "true" and "false", but not totally case insensitive. maybe:
```
if data in ("true", "True", "TRUE", "on", "On", "ON"):
    return True
elif data in ("false", "False", "FALSE", "off", "Off", "OFF"):
    return False
elif ...
```

> Also isdigit handles positive ints only.
good catch! would prefer not to use regex but so it must be

JerrySentry (approved):

PR #53, opened 2023-09-28
chore: Handle gh refresh attempt failure more gracefully

giovanni-guidini:
Instead of throwing error we're gonna pretend like it didn't happen.

### Legal Boilerplate

### DIFF
--- a/shared/torngit/github.py
+++ b/shared/torngit/github.py
@@ -331,9 +331,17 @@ class Github(TorngitBaseAdapter):
                 }
             )
             return self.token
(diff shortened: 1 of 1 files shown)


matt-codecov (approved):
gio and i talked about the lock-based refresh approach described on https://github.com/codecov/shared/pull/52. he's going to cut an issue for that refactor and explain some constraints i didn't cover and we can work it onto a sprint

gio said for now we should merge this log and make the error less disruptive, and if we can get to the token refresh refactor fast enough then we don't need to do the "reload token from db" approach that was started in https://github.com/codecov/shared/pull/52

PR #57, opened 2023-09-29
implement marketo

dana-yaish:
We need to integrate marketo and send it events, just like we do with pubsub and segment 
most of the logic is taken from [sentry](https://l.codecov.dev/tn8BBH)

### Legal Boilerplate


shared/analytics_tracking/marketo.py
```diff
@@ -0,0 +1,71 @@
...
+    def track_event(self, event: Event, *, is_enterprise, context: None):
+        body = {
+            "input": [event.serialize()],
+        }
+        return self.make_rest_request(self.LEAD_URL, method="POST", json=body)
+
+    def make_request(self, url, *args, **kwargs):
+        full_url = self.base_url + url
+        method = kwargs.pop("method", "GET")
+        res = self.client.request(method, full_url, **kwargs)
```
giovanni-guidini, line 47: The one observation I'd like to make is to maybe add a `try` `except https.HTTPError` around the `self.client.request` method in case the _request_ fails.

And like just emit a warning or something. To prevent this from blowing up.
But chances of this happening are pretty low I suppose...

giovanni-guidini (approved):
LGTM. Left 1 comment.

shared/analytics_tracking/marketo.py
dana-yaish: we already catch any exceptions that might happen in manager.py see [here](https://github.com/codecov/shared/blob/main/shared/analytics_tracking/manager.py#L34-L40). I'll add a catch If you think it's worth adding it here too

PR #72, opened 2023-11-02
add django/django-admin reqs, create timeseries django models

matt-codecov:
https://github.com/codecov/engineering-team/issues/547
https://github.com/codecov/worker/pull/168

this PR copies some django stuff from `codecov-api` so we can share it with `worker`:
- `DatabaseRouter` (renamed to `MultiDatabaseRouter` in `shared.django_apps.db_routers`)
- database settings (found in `shared.django_apps.db_settings`)

it also adds some new django stuff:
- `pg_telemetry`, an app that defines a postgres-backed model for timeseries data
- `ts_telemetry`, an app that defines a timescale-backed model for timeseries data
- `TelemetryDatabaseRouter`, a router that points the telemetry apps at their respective databases until we pick one
- dummy config, tests, migrations

`pg_telemetry` and `ts_telemetry` each define a `SimpleMetric` model which logs a (name, value) pair to `telemetry_simple` in their respective databases. `pg_telemetry` records can be tagged with `repo_id`, `owner_id`, and `commit_id`, which are foreign keys to those tables. `ts_telemetry` cannot join across databases so for usability's sake its records can be tagged with `repo_slug`, `owner_slug`, and `commit_slug`. i'll introduce usages in `worker` in another PR

patch coverage on this PR will be low because it doesn't super make sense to cover `dummy_settings.py`, `db_settings.py`, `manage.py` but i added tests for the database routers which the api didn't have which should win me brownie points
 
### Legal Boilerplate


giovanni-guidini (approved):
I read through the code and it seems fine. The changes make sense, but idk all that much about Django configuration.

PR #73, opened 2023-11-03
replacing newheader and newfooter with condensed_header and condensed_footer

dana-yaish:
since newheader, newfooter and newfiles will be outdated names soon, we'll be using condensed_header, condensed_footer and condensed_files instead with keeping the old names valid for backward compatibility 

### Legal Boilerplate


giovanni-guidini (approved):
Please add a comment next to "newfooter", "newheader" etc mentioning that it's deprecated and they are there for backwards compatibility, etc.

Basically what you say in the PR comment.

PR #74, opened 2023-11-06
fix: Add report size method

scott-codecov:
Once incorporated into worker can be used to fix https://codecov.sentry.io/issues/4597918964/

### DIFF
--- a/shared/reports/resources.py
+++ b/shared/reports/resources.py
@@ -633,6 +633,13 @@ class Report(object):
     def from_chunks(cls, *args, **kwargs):
         return cls(*args, **kwargs)
 
+    @property
+    def size(self):
+        size = 0
+        for chunk in self._chunks:
+            size += len(chunk)
+        return size
+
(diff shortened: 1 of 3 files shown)


shared/reports/resources.py
```diff
@@ -633,6 +633,13 @@ def __init__(
     def from_chunks(cls, *args, **kwargs):
         return cls(*args, **kwargs)
 
+    @property
+    def size(self):
```
giovanni-guidini, line 637: What's the unit? Is it... number of lines? Size of chunks in bytes? If you know please add a comment with that info

giovanni-guidini (approved):
LGTM. Small comment.

shared/reports/resources.py
scott-codecov: It's kind of a unitless number.  It's the summed number of characters in each chunk - and the chunk itself is newline-separated JSON.  So sorta bytes but not really?

PR #75, opened 2023-11-06
fix: make sure EditableReport is merged with EditableReportFile

giovanni-guidini:
When processing reports (more than 1 per `UploadProcessor` task) it's possible
that we will merge a `Report` into a `EditableReport` (the original).
This is fine, but we were not respecting the fact that files inside
`EditableReport` should be `EditableReportFile`.

These changes force all files merged into an `EditableReport` to be
of type `EditableReportFile`. Merge is left unchanged, cause it seems
complex. Instead we just run an extra check and cast to `EditableReportFile`
if needed.

closes: codecov/internal-issues#nternal-issues#138

### Legal Boilerplate

### DIFF
--- a/shared/reports/editable.py
+++ b/shared/reports/editable.py
@@ -14,14 +14,30 @@ class EditableReportFile(ReportFile):
 
     __slots__ = ("_details",)
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    @classmethod
+    def from_ReportFile(cls, report_file: ReportFile):
+        name = report_file.name
+        editable_file = cls(name)
+        editable_file._totals = report_file._totals
+        editable_file._session_totals = report_file._session_totals
+        editable_file._lines = report_file._lines
+        editable_file._line_modifier = report_file._line_modifier
+        editable_file._ignore = report_file._ignore
+        editable_file._details = report_file._details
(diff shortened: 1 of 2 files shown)


shared/reports/editable.py
```diff
@@ -14,15 +14,31 @@ class EditableReportFile(ReportFile):
...
+    def from_ReportFile(cls, report_file: ReportFile):
+        name = report_file.name
+        editable_file = cls(name)
+        editable_file._totals = report_file._totals
+        editable_file._session_totals = report_file._session_totals
+        editable_file._lines = report_file._lines
+        editable_file._line_modifier = report_file._line_modifier
+        editable_file._ignore = report_file._ignore
+        editable_file._details = report_file._details
+        editable_file.fix_details()
```
scott-codecov, line 27: This already happened in the `__init__` method when calling `cls(name)` above.  Do we need to do it again here?

scott-codecov (approved):

shared/reports/editable.py
scott-codecov: Oh I think the answer is "yes" since we're setting `_details` right above this line.

giovanni-guidini: Exactly, I think we do need to do that again.

PR #78, opened 2023-11-13
ref: Include GitHub app token code from worker

scott-codecov:
We need the `get_github_integration_token` in `codecov-api`

No new code here - just pulled things from the following `worker` sources:

* https://github.com/codecov/worker/blob/main/services/github.py
* https://github.com/codecov/worker/blob/main/services/pem.py
* https://github.com/codecov/worker/blob/main/services/tests/test_github.py

### DIFF
--- a/shared/github/__init__.py
+++ b/shared/github/__init__.py
@@ -1,0 +1,88 @@
+import logging
+from datetime import datetime
+from time import time
+from typing import Optional
+
+import jwt
+import requests
+
+import shared.torngit as torngit
+from shared.config import get_config, load_file_from_path_at_config
+
+log = logging.getLogger(__name__)
+
+loaded_pems = None
+
+pem_paths = {
+    "github": ("github", "integration", "pem"),
+    "github_enterprise": ("github_enterprise", "integration", "pem"),
+}
+
+
+def get_pem(pem_name: str) -> str:
+    path = pem_paths[pem_name]
+    return load_file_from_path_at_config(*path)
+
+
+class InvalidInstallationError(Exception):
+    pass
+
+
+def get_github_integration_token(service, integration_id=None) -> Optional[str]:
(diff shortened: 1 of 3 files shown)


giovanni-guidini:
The `cache` had been moved already. It was duplicated in the worker: https://github.com/codecov/shared/blob/gio/label-lookup/shared/helpers/cache.py#L1

shared/github/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+        if res.status_code in (404, 403):
+            log.warning(
+                "Integration could not be found to fetch token from or unauthorized",
+                extra=dict(
+                    git_service=service,
+                    integration_id=integration_id,
+                    api_endpoint=api_endpoint,
+                ),
+            )
+            return None
```
scott-codecov, line 63: Oh, this is a code change.  Code in worker raises a custom exception here.

giovanni-guidini: Hum... should we raise something here as well?

shared/helpers/cache.py
```diff
@@ -269,3 +269,6 @@ async def wrapped(*args, **kwargs):
             return result
 
         return wrapped
+
+
+cache = OurOwnCache()
```
giovanni-guidini, line 274: I actually think you should put this in a file in the worker... but this is a personal and nit opinion.

scott-codecov: Oh yeah, I guess I can leave the instance in `worker` and `codecov-api` and then just wrap the `shared` function there (but with caching)

shared/github/__init__.py
scott-codecov: Yeah I guess I can create a custom exception `shared` and then catch it in `worker` and re-raise the other custom exception.

giovanni-guidini (approved):

PR #79, opened 2023-11-13
Change "label" in report by "label_id" and have a lookup table

giovanni-guidini:
# Context
This is a draft pull request, but it's close to what will probably be the final solution.
"label index" and "label ID" are used interchangeably.

Recently we had issues with some large customer using label analysis because the report with labels simply uses too much space. For that particular customer the chunks size was more than 1.5Gb and so worker can't load it into memory. This happens because the labels are saved in their canonical format in `CoverageDatapoint.labels`. These are moderately long strings that are repeated all through the file.

# Solution
To unblock customers wanting to try ATS we will be encoding them. The idea is that the `CoverageDatapoints` will only keep label IDs (which are much smaller than the labels themselves) and `Report` will have a "detachable" label index table.

Detachable in this case means it will live in a separate file (not `chunks.txt`) and we will be able to load and unload it into memory as needed, to further save space.

Notice that it is technically possible to operate over the IDs, so typically we only need to know what the actual labels are when merging a new report and during the `label_analysis` task. That is partly the reason that only the `Report` (top level) has access to the label index table.

# Comments on design choices
The other part is that I'd like to avoid inconsistencies in the index. So the `CoverageDatapoints` shouldn't be able to modify the index at all. In fact for the `CoverageDatapoint` it doesn't matter what the label is, so it can be an index and that's ok.
I purposely didn't add functions to alter the label index table to the report. I think it's better to be more explicit about it and let the worker do that.

So in the case of `Report` all you can do is load or unload a label index table, and look up labels by ID. It panics if it can't find a label, and that's it.
The worker code will make sure that we are using it in the expected way.

The other intentional choice was changing `CoverageDatapoint.labels` to `CoverageDatapoint.label_ids`. I think it's important that the name represents what it actually is. Specially from a maintenance perspective, it's a change that I think it's worth doing now.

Yes it increases the chance of bugs, but we do want things to not fail silently in this case, I'd argue. This is a complex change (when you factor in the worker piece), and the `Report` class is very important for the core product, so it's better to catch bugs fast during testing. Increasing the likelihood of loud bugs and exceptions being raised in this case is actually a good thing, because we can fix the data flows that we missed fast when testing in staging.

### DIFF
--- a/shared/reports/editable.py
+++ b/shared/reports/editable.py
@@ -49,17 +49,26 @@ class EditableReportFile(ReportFile):
         res["present_sessions"] = sorted(self._details.get("present_sessions"))
         return res
 
-    def delete_labels(self, session_ids_to_delete, labels_to_delete):
+    def delete_labels(
+        self, session_ids_to_delete: List[int], label_ids_to_delete: List[int]
+    ):
+        """Given a list of session_ids and label_ids to delete
+        Remove all datapoints that belong to at least 1 session_ids to delete and include
+        at least 1 of the label_ids to be removed
+        """
         for index, line in self.lines:
             if line.datapoints is not None:
                 if any(
-                    (dp.sessionid in session_ids_to_delete and lb in labels_to_delete)
+                    (
+                        dp.sessionid in session_ids_to_delete
+                        and label_id in label_ids_to_delete
+                    )
                     for dp in line.datapoints
-                    for lb in dp.labels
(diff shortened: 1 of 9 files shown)


joseph-sentry (approved):

scott-codecov (approved):

matt-codecov (approved):

giovanni-guidini:
This PR is being replaced.

PR #84, opened 2023-11-16
fix: Timeout Github API student check (#277)

JerrySentry:
Add a short timeout to the GitHub education API we use to check if a user is student. This GH call is known to be flaky, causing timeouts when this is called during the login step. This change gives it 3s to respond before default to "not a student" like how it is being handled currently when there's other errors coming from this GH call.

Note that after this is released and API is using this package, we need to turn off the `student_disabled` config in API.

### Legal Boilerplate

### DIFF
--- a/shared/torngit/base.py
+++ b/shared/torngit/base.py
@@ -50,8 +50,11 @@ class TorngitBaseAdapter(object):
         "xtend",
     )
 
-    def get_client(self):
-        timeout = httpx.Timeout(self._timeouts[1], connect=self._timeouts[0])
+    def get_client(self, timeouts: List[int] = []) -> httpx.AsyncClient:
+        if timeouts:
+            timeout = httpx.Timeout(timeouts[1], connect=timeouts[0])
+        else:
+            timeout = httpx.Timeout(self._timeouts[1], connect=self._timeouts[0])
         return httpx.AsyncClient(
             verify=self.verify_ssl
             if not isinstance(self.verify_ssl, bool)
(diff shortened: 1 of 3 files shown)


shared/torngit/base.py
```diff
@@ -50,8 +50,11 @@ class TorngitBaseAdapter(object):
         "xtend",
     )
 
-    def get_client(self):
-        timeout = httpx.Timeout(self._timeouts[1], connect=self._timeouts[0])
+    def get_client(self, timeouts: List[int] = []) -> httpx.AsyncClient:
+        if timeouts:
+            timeout = httpx.Timeout(timeouts[1], connect=timeouts[0])
+        else:
+            timeout = httpx.Timeout(self._timeouts[1], connect=self._timeouts[0])
```
rohitvinnakota-codecov, line 57: Was this call previously defaulting to `[10,30]` for its timeout? JC what is the advantage of switching over to a 3 second timeout in this case?

JerrySentry: Yeah, its 30s timeout to hear back from the response (10s timeout to connect), so when it flakes, we hang for 30s in our login request and then the timeout is not handled so the client call to us breaks. This change makes it such that we only give GH 3s to respond and it will catch timeouts and handle the exception (default to not a student). This way the login doesn't hang and break, instead it will wait a bit and still login

rohitvinnakota-codecov: Awesome, thanks for the explanation.

PR #86, opened 2023-11-17
Adding host override tech

trent-codecov:
### Legal Boilerplate

### DIFF
--- a/shared/torngit/github.py
+++ b/shared/torngit/github.py
@@ -71,7 +71,7 @@ class Github(TorngitBaseAdapter):
 
     @classmethod
     def get_service_url(cls):
-        return get_config("github", "url", "https://github.com").strip("/")
+        return get_config("github", "url", default="https://github.com").strip("/")
 
     @property
     def service_url(self):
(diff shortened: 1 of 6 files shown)


shared/torngit/github.py
```diff
@@ -160,6 +181,11 @@ async def make_http_call(
 
         url = url_concat(url, args).replace(" ", "%20")
 
+        if url.startswith(self.api_url) and self.api_host_header is not None:
+            _headers["Host"] = self.api_host_header
+        elif url.startswith(self.api_url) and self.host_header is not None:
```
scott-codecov, line 186: Should this be `self.service_url` here?

trent-codecov: Yep. Fixed!

shared/torngit/github.py
```diff
@@ -70,10 +69,44 @@ class Github(TorngitBaseAdapter):
...
+            "/"
+        )
+
+    @classmethod
+    def get_api_host_header(cls):
+        return get_config("github", "api_host_override")
+
+    @classmethod
+    def get_host_header(cls):
+        return get_config("github", "host_override")
```
scott-codecov, line 92: You could do `get_config(cls.service, "host_override")` here and then remove the similar method in `github_enterprise.py`  (same for `get_api_host_header` above)

scott-codecov (approved):

shared/torngit/github.py
trent-codecov: Thanks for the tip. Added

PR #90, opened 2023-11-30
feat: Add core bundle analysis report handling

scott-codecov:
Some preliminary work related to bundle analysis.

This introduces some of the core classes that will represent a bundle analysis report.  We will ingest bundle stats files (JSON) and store the data inside a SQLite database which will live in GCS.  When we need to merge more bundle stats data into the report we'll download the SQLite file from storage, ingest the new stats JSON and the re-upload the SQLite file back to storage.  The goal of using SQLite here is to keep our memory footprint low.  We're using a streaming JSON parser and write data directly into SQLite as each piece is parsed (buffering only tiny bits of data in-memory as needed).

Example usage: ingesting a new stats JSON file (in worker)

```python
from shared.storage import get_appropriate_storage_service
from shared.bundle_analysis import BundleReport, BundleReportLoader


storage_service = get_appropriate_storage_service()

bundle_loader = BundleReportLoader(storage_service, repo_hash)
bundle_report = bundle_loader.load(commit_sha)
if bundle_report is None:
    bundle_report = BundleReport()

stats_json_path = download_bundle_stats_from_storage(upload)
bundle_report.ingest(stats_json_path)
bundle_loader.save(bundle_report, commit_sha)
bundle_report.cleanup()
```

I don't really know everything we'll need in terms of APIs from these classes but took a few educated guesses inside `BundleReport` and `BundleAnalysisComparison`.

For more context on bundle analysis in general please see:
* https://www.notion.so/sentry/Bundle-Analysis-in-JavaScript-124dfedf5ced4cb2aa0cf7c548c61496
* https://www.notion.so/sentry/Codecov-Bundle-Analysis-Engineering-Spec-2a75e5664361413ca1fc8c5c6113b445
* https://www.notion.so/sentry/Bundle-Analysis-Plugin-b4d5dc2cf0a24ab3a225cc08704c3a25

If you're concerned about SQLite file size I think it should be OK.  Here's some tests with synthetic data: 
Screenshot 2023-11-28 at 12 53 32 PM

Need to do some more testing on some real (and very large) stats JSON files.


shared/bundle_analysis/report.py
```diff
@@ -0,0 +1,122 @@
...
+from sqlalchemy.exc import OperationalError
+from sqlalchemy.orm import Session
+from sqlalchemy.sql import func
+
+from . import models
+from .parser import parse
+
+# matches file paths with content hashes like:
+# path/to/file-aa21a697.ext
+CONTENT_HASH_RE = re.compile(r"^.+(-[\w\d]{8})\.[\w\d]+$")
```
scott-codecov, line 16: Abhi pointed me to https://github.com/getsentry/relay/blob/24b39d09d7f59f95c3cbb37fed0839ce29e33ec3/relay-event-normalization/src/regexes.rs#L66 which looks a bit more comprehensive in terms of the hash part of the regex

shared/bundle_analysis/models.py
```diff
@@ -0,0 +1,169 @@
...
+
+def get_db_session(path: str) -> DbSession:
+    engine = create_engine(f"sqlite:///{path}")
+    Session = sessionmaker()
+    Session.configure(bind=engine)
+    session = Session()
+    return session
+
+
+assets_chunks = Table(
```
giovanni-guidini, line 75: Very crafty the way you declared the tables that glue together N:N relationships without having to create new models just for that. I like it.

However it's not immediately obvious why these declarations are different from the rest (at least not to me), and that you don't have to use these as models.

I think a comment before the `Table` declarations explaining what they do and that it doesn't have to be a model in itself but needs to exist in the DB would be helpful here.

giovanni-guidini (commented):
Looking good. I especially like all the typings. Very helpful 🙏, thanks 
Left some comments.

shared/bundle_analysis/__init__.py
```diff
@@ -0,0 +1,10 @@
+from . import models
```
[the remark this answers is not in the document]
scott-codecov: No reason - I can switch these to full module names instead

shared/bundle_analysis/parser.py
```diff
@@ -0,0 +1,139 @@
...
+        asset = None
+        chunk = None
+        chunk_asset_names = []
+        module = None
+        module_chunk_external_ids = []
+
+        for (prefix, event, value) in events:
+            # assets
+            if (prefix, event) == ("assets.item", "start_map"):
+                # new asset
```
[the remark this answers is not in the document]
giovanni-guidini: That is a possibility. Or you can create just a data class for the `ParserState` and pass that around to be mutated. Whatever you prefer.

shared/bundle_analysis/report.py
scott-codecov: This sort of normalization is now happening in the JS plugin that generates the stats JSON

giovanni-guidini (approved):

PR #94, opened 2023-12-04
Create `Report.header` and populate it with labels_index

giovanni-guidini:
Creates a `Report.header` with information about the report (similar to `ReportFile.details`). Populates it with the labels_index.

Adds the util functions to get/set the header, encode it into the `chunks.txt`, and build reports with it.

Also taking opportunity for a small optimization around labels.


giovanni-guidini:
These changes are meant to replace https://github.com/codecov/shared/pull/79

scott-codecov (commented):
Nice, I think this makes a lot of sense.  It felt weird the other way where reads from the storage service were sort of interleaved with core report behavior.  I think this approach will lead to a much cleaner and understandable implementation.

shared/utils/merge.py
```diff
@@ -201,9 +201,17 @@ def merge_datapoints(
 ):
     if d1 is None and d2 is None:
         return None
+    # Remove duplicates
+    # str(dp) -> dp
+    index_of_dps = dict()
+    both_lists = filter(None, (d1 or []) + (d2 or []))
+    for dp in both_lists:
+        key = str(dp)
```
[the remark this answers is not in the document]
giovanni-guidini: `dp` is not a label_id. It's a `CoverageDatapoint`.

shared/reports/resources.py
```diff
@@ -587,6 +587,17 @@ def line_without_multiple_sessions(
         )
 
 
+def chunks_from_storage_contains_header(chunks: str) -> bool:
```
scott-codecov, line 590: This looks good as long as the header JSON never contains newlines

scott-codecov (approved):

shared/reports/resources.py
giovanni-guidini: hum... fair point. Labels can get pretty wild. I think we should assume that the labels in the header might contain newlines.

Do you have any suggestions for the function?

giovanni-guidini: So I did test that and it seems we're fine even if the labels have "\n" in them cause they'll be JSON stringified... so I guess it's fine (at least for now) to leave as is.

```python
>>> d = {"strange_label": "big label with \n \n in it"}
>>> print(d["strange_label"])
big label with 
 
 in it
>>> x = json.dumps(d)
>>> x
'{"strange_label": "big label with \\n \\n in it"}'
>>> "\n" in x
False
>>> x.index("\n")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: substring not found
>>> x.index('\n')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: substring not found
>>> x.split("\n")
['{"strange_label": "big label with \\n \\n in it"}']

Comment (giovanni-guidini) [reply shared/reports/resources.py:590]: The alternative

```python
def chunks_from_storage_contains_header(chunks: str) -> bool:
    possible_first_chunk, _ = chunks.split(END_OF_CHUNK, 1)
    return END_OF_HEADER in possible_first_chunk
```

is most likely less fast that the current implementation. Cause the `possible_first_chunk` might be pretty big.

PR #95, opened 2023-12-04
Create upload test tasks

joseph-sentry:
Depends on: #91

### DIFF
--- a/shared/utils/enums.py
+++ b/shared/utils/enums.py
@@ -43,4 +43,5 @@ class TaskConfigGroup(Enum):
     synchronize = "synchronize"
     timeseries = "timeseries"
     upload = "upload"
+    upload_test = "upload_test"
     verify_bot = "verify_bot"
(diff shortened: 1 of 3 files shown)


giovanni-guidini (commented):
I think calling it just "test" will cause confusion because it's too generic of a name. To make sure it's clear that this is part of test result processing product I think you should always refer to it as "test result" or "test outcome".

shared/utils/enums.py
```diff
@@ -43,4 +43,5 @@ class TaskConfigGroup(Enum):
     synchronize = "synchronize"
     timeseries = "timeseries"
     upload = "upload"
+    upload_test_result = "upload_test_result"
```
scott-codecov, line 46: Maybe just `test_results`?  All the test result ingestion tasks can be in that same group.  I'm using `app.tasks.bundle_analysis.<TaskName>` for bundle analysis stuff.

PR #99, opened 2023-12-12
ref: Store bundle analysis reports per report key

scott-codecov:
Tiny change here - just renaming things.  We could technically have multiple bundle reports per commit (via `report_code`) so this naming and storage layout makes more sense.  `report_key` will be a `CommitReport`'s `external_id` UUID in practice.


shared/bundle_analysis/comparison.py
```diff
@@ -40,21 +40,23 @@ class BundleAnalysisComparison:
     Compares bundle reports on two different commits.
     """
 
-    def __init__(self, loader: BundleReportLoader, base_sha: str, head_sha: str):
+    def __init__(
+        self, loader: BundleReportLoader, base_report_key: str, head_report_key: str
```
giovanni-guidini, line 44: You mention
> We could technically have multiple bundle reports per commit (via report_code)

The default `report_code` is `None`. In this case the `base_report_key` will be `"None"` or you'll be passing the commit sha to be the default key?

I suppose it's something the processing code will take core of, and it doesn't really matter. I'm just curious 👀

giovanni-guidini (commented):
Changes look fine ✅ very clean and easy to follow along.

I wonder though if it makes sense to have the bundle path be a function of `commit_sha` AND `report_key`? Is that something you've considered but discarded due to extra complexity or something?

shared/bundle_analysis/comparison.py
scott-codecov: The report key could be something like `commit_report.external_id` which seems to always be a UUID

giovanni-guidini: I like that... cool cool

shared/bundle_analysis/storage.py
```diff
@@ -11,7 +11,7 @@
 
 
 class StoragePaths(Enum):
-    bundle_report = "v1/repos/{repo_key}/commits/{commit_sha}/bundle_report.sqlite"
```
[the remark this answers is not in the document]
scott-codecov: I suppose storing them under the commit could be useful if you were just browsing around in the GCS explorer?  I kind of wanted to keep this part fairly generic.  The application code could make the `report_key` something like `{commit_report.commit.commit_id}/{commit_report.external_id}` if desired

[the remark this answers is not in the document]
giovanni-guidini: hum I don't think that's desired.... keeping the key simple seems the optimal solution I guess.

It's ok, we can probably find the path in some record in our DB. That's fine.

giovanni-guidini (approved):

PR #100, opened 2023-12-13
feat: Support multiple bundles per bundle analysis report

scott-codecov:
Support multiple bundles per bundle analysis report.  A bundle is a high level thing that might exist in a monorepo (i.e. something like "dashboard" or "admin" etc.).  A single session/upload contains information pertaining to a single bundle.  See the schema changes in `models.py` for a more detailed understanding of the relationships.

### DIFF
--- a/shared/bundle_analysis/report.py
+++ b/shared/bundle_analysis/report.py
@@ -8,10 +8,14 @@ from sqlalchemy.orm import Session as SQLAlchemySession
 from sqlalchemy.sql import func
 
 from shared.bundle_analysis import models
-from shared.bundle_analysis.parser import parse
+from shared.bundle_analysis.parser import Parser
 
 
-class Bundle:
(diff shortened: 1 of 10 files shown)


shared/bundle_analysis/report.py
```diff
@@ -80,27 +122,33 @@ def cleanup(self):
         self.db_session.close()
         os.unlink(self.db_path)
 
-    def ingest(self, path: str) -> int:
+    def ingest(self, path: str, bundle_name: Optional[str] = None) -> int:
```
scott-codecov, line 125: TODO: not sure yet if this will be passed in or parsed directly out of the JSON

giovanni-guidini (approved):
LGTM, but I borrowed heavily from my mental model about the coverage data 👀 (meaning I might have overlooked / not understood details)

PR #103, opened 2024-01-03
tests: update test_install_validation to check for warnings

joseph-sentry:
Fixes: https://github.com/codecov/engineering-team/issues/872


giovanni-guidini (approved):
Thanks for the thorough work 🙏

PR #105, opened 2024-01-08
fix: also change regex for flag names in flag_list_structure

joseph-sentry:


matt-codecov (approved):
are there tests to update for this?

PR #110, opened 2024-01-22
feat: Fix bundle analysis DB model and upload parsing

JerrySentry:
- Allow parsing of multiple uploads where there's duplicate chunk IDs between uploads
- Allow re-uploading a bundle file where it has the same bundle name, the previous upload will be fully replaced by the current one

https://github.com/codecov/engineering-team/issues/1079
https://github.com/codecov/engineering-team/issues/1080

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/models.py
+++ b/shared/bundle_analysis/models.py
@@ -43,11 +43,12 @@ create index assets_name_index on assets (name);
 
 create table chunks (
     id integer primary key,
+    session_id integer not null,
     external_id text not null,
     unique_external_id text not null,
     entry boolean not null,
     initial boolean not null,
-    unique (unique_external_id)
+    foreign key (session_id) references sessions (id)
 );
 
 create table assets_chunks (
@@ -60,8 +61,10 @@ create table assets_chunks (
 
 create table modules (
     id integer primary key,
+    session_id integer not null,
     name text not null,
-    size integer not null
+    size integer not null,
+    foreign key (session_id) references sessions (id)
 );
 
 create table chunks_modules (
@@ -160,7 +163,9 @@ class Asset(Base):
     size = Column(types.Integer, nullable=False)
 
     session = relationship("Session", backref=backref("assets"))
-    chunks = relationship("Chunk", secondary=assets_chunks, back_populates="assets")
+    chunks = relationship(
+        "Chunk", secondary=assets_chunks, back_populates="assets", cascade="all, delete"
+    )
 
 
 class Chunk(Base):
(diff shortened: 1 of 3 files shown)


shared/bundle_analysis/models.py
```diff
@@ -43,11 +43,12 @@
```
JerrySentry, line 43: Add a session_id FK for Chunk and Modules, as each Chunk and Module can only belong to one session (ie stat file / bundle report). This way we can quickly navigate to the right objects of a session and it will fix an issue where Chunks/Modules from other bundles are incorrectly associated with the current bundle being processed.

shared/bundle_analysis/models.py
```diff
@@ -43,11 +43,12 @@
 
 create table chunks (
     id integer primary key,
+    session_id integer not null,
     external_id text not null,
     unique_external_id text not null,
     entry boolean not null,
     initial boolean not null,
-    unique (unique_external_id)
```
JerrySentry, line 50: Lifting this uniqueness constraint. Discussed with BA folks, that this shouldn't be here to begin with.

shared/bundle_analysis/parser.py
```diff
@@ -87,8 +105,8 @@ def _parse_event(self, event: Tuple[str, str, str]):
             if bundle is None:
                 bundle = Bundle(name=value)
                 self.db_session.add(bundle)
```
JerrySentry, line 107: Fixes a bug where only save the bundle's new session if the bundle doesn't exist (ie new upload)

shared/bundle_analysis/parser.py
```diff
@@ -174,7 +192,10 @@ def _create_associations(self):
         # associate chunks to assets
         inserts = []
         for chunk_id, asset_names in self.chunk_asset_names_index.items():
```
JerrySentry, line 194: Only build associations of modules/chunks within the same session, when other bundle reports have the same ID/name it would ignore it as each bundle file is independent with one another

shared/bundle_analysis/parser.py
```diff
@@ -62,6 +62,24 @@ def parse(self, path: str) -> int:
...
+            # Delete old session/asset/chunk/module with the same bundle name if applicable
+            old_session = (
+                self.db_session.query(Session)
+                .filter(
+                    Session.bundle == self.session.bundle, Session.id != self.session.id
+                )
+                .one_or_none()
+            )
+            if old_session:
+                for model in [Asset, Chunk, Module]:
```
adrian-codecov, line 74: With all the cascading you set before, would this work if you only deleted the session?

adrian-codecov (commented):
Did have that one qs left there but overall lgtm. I'll approve but lets merge after we figure do the E2E ya?

adrian-codecov (approved):

shared/bundle_analysis/parser.py
JerrySentry: I would need to do a  `cascade="all, delete"` for Assets/Chunks/Modules as a relationship on the Sessions table, then when I delete the session it should delete the related ACM objects, then the other many-to-many table objects.
But for some reason add the `relationship` line in the Session model doesn't work, I've spent a decent amount of time on it before giving up :(

PR #112, opened 2024-01-29
build: upload test results

joseph-sentry:
Uploads test results to codecov

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -204,7 +204,7 @@ jobs:
           pip install -r tests/requirements.txt
           pip install -r requirements.txt
           python setup.py develop
-          pip install codecov-cli>=0.4.6
+          pip install 'codecov-cli>=0.4.6'
 
       - name: Run tests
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -204,37 +204,41 @@ jobs:
...
           pip install -r requirements.txt
           python setup.py develop
-          pip install codecov-cli
+          pip install 'codecov-cli>=0.4.6'
 
       - name: Run tests
         run: |
           . venv/bin/activate
-          python -m pytest --cov=./
+          python -m pytest --cov=./ --junitxml=junit.xml
```
adrian-codecov, line 212: Thoughts of using the workflow you did in this PR, https://github.com/codecov/gha-workflows/pull/18/files, so that it's the same across all repos and you don't have to change it in multiple places? Like here, https://github.com/codecov/worker/blob/main/.github/workflows/ci.yml#L24

adrian-codecov (approved):

PR #120, opened 2024-02-06
feat: get repos from graphql

giovanni-guidini:
context: codecov/engineering-team#971

Adds function to get repositories from Github's GraphQL api.

It uses node_id to get the specified node from gh. This can be found in
the INSTALLATION webhook that we will process.
Along with it there will be the databaseId (service_id).

Adding the `expected_owner` entry because when processing an installation webhook
we do know what the expected owner is. Fetching the owner would require another request,
but can be done (again by node_id)

The idea is to use this function when syncing repos if we know what the repos
covered by an installation are.

### DIFF
--- a/shared/torngit/github.py
+++ b/shared/torngit/github.py
@@ -36,7 +36,10 @@ log = logging.getLogger(__name__)
 
 METRICS_PREFIX = "services.torngit.github"
 
-GITHUB_REPO_COUNT_QUERY = """
+
+class GitHubGraphQLQueries(object):
+    _queries = dict(
+        REPO_TOTALCOUNT="""
 query {
     viewer {
         repositories(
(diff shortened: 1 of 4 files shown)


shared/torngit/github.py
```diff
@@ -47,11 +50,46 @@
         }
     }
 }
-"""
+""",
+        REPOS_FROM_NODEIDS="""
```
matt-codecov, line 54: does github let you fetch the whole list without pagination or will it paginate for you with some default page size?

https://github.com/codecov/engineering-team/issues/139 has an example of pagination with github's graphql api if we need to add it

tests/integration/cassetes/test_github/TestGithubTestCase/test_get_repos_from_nodeids_generator.yaml
```diff
@@ -0,0 +1,81 @@
+interactions:
+- request:
+    body: '{"query": "\nquery GetReposFromNodeIds($node_ids: [ID!]!) {\n    nodes(ids:
+      $node_ids) {\n        __typename \n        ... on Repository {\n            #
+      databaseId == service_id\n            databaseId\n            name\n            primaryLanguage
+      {\n                name\n            }\n            isPrivate\n            defaultBranchRef
+      {\n                name\n            }\n            owner {\n                #
+      This ID is actually the node_id, not the ownerid\n                id\n                login\n            }\n        }\n    }\n}\n",
```
matt-codecov, line 8: lol it includes our comments

shared/torngit/github.py
giovanni-guidini: That's a very good question that I probably should have put in the PR description.
I spent quite some time trying to get pagination in this query. From the [docs](https://github.com/codecov/codecov-cli/pull/373) it indicates you can specify a `first`, `last` and get `pageInfo` cursors.

But from the explorer docs the `nodes` query doesn't include the inputs for `first` and `last` (I tried, didn't work) AND the `Repostory` object doesn't include `pageInfo` as a possible field to query. For that reason I don't _think_ there's pagination? At least I couldn't find a way to paginate the request.


Screenshot 2024-02-07 at 09 08 46

giovanni-guidini: update: I decided to implement simple pagination from our side just in case.

shared/torngit/github.py
```diff
@@ -47,11 +50,61 @@
...
+            }
+            isPrivate
+            defaultBranchRef {
+                name
+            }
+            owner {
+                # This ID is actually the node_id, not the ownerid
+                id
+                login
+            }
```
matt-codecov, line 73: do we need a separate query for getting owner information or could we include databaseId here?

matt-codecov (commented):
code change looks good to me, would be nice not to need an extra query if we can work that owner info into the first one

shared/torngit/github.py
giovanni-guidini: I would have included if it was possible

matt-codecov: RIP

matt-codecov (approved):

PR #124, opened 2024-02-12
fix: Retrieve large source files from GH properly

JerrySentry:
When source files are larger than 1MB, we need to call another URL to retrieve its contents

### Legal Boilerplate


shared/torngit/github.py
```diff
@@ -1105,14 +1105,29 @@ async def get_source(self, path, ref, token=None):
                     ref=ref,
                     token=token,
                 )
+
+                # When file size is greated than 1MB, content would not populate,
```
ajay-sentry, line 1109: nit: typo "greater"

shared/torngit/github.py
```diff
@@ -1105,14 +1105,29 @@ async def get_source(self, path, ref, token=None):
...
+                # instead we have to retrieve it from the download_url
+                if (
+                    not content.get("content")
+                    and content.get("download_url")
+                    and content.get("encoding") == "none"
```
ajay-sentry, line 1114: Naive question here, is "none" a default/standard value? Would it make sense to make this a const?

JerrySentry: "none" is the value GH API sends when there is no `content` value in the response. The `content` field is basically the blob of source code and `encoding` is how the blob is encoded. I don't think we need a const here, this logic is just a one-off and the encoding response won't be handled anywhere else in the future.

rohitvinnakota-codecov (approved):

PR #127, opened 2024-02-19
ci: Build only in python 3.12

giovanni-guidini:
context: codecov/engineering-team#874

Changing `shared` default image to use Python 3.12. We want to upgrade to the latest version.
Same move will be done in the `worker` and `api` repos as well.

Disabling ATS (not being used) and `mutmut` (doesn't work with 3.12 / not being used)

ps.: Obviously we need to update worker and api before using a version of shared that requires 3.12

UPDATE
Now that worker and api have been updated to python 3.12 we can update this one too


giovanni-guidini:
About the `Run Mutations` step.
It seems that `mutmut` runs (as in the mutants are executed).

The command that actually fails is the report generation (`mutmut junitxml > /tmp/artifacts/mut.xml`)... I tested locally (reproduced the same error). Also `mutmut show all` doesn't work (different error). Not sure why...

joseph-sentry (approved):
lgtm

just to make sure, this will mean we can't update the version of shared that we use to a version after this one, until the `worker` and `api` are both updated to use `3.12`?

Do we think there will be issues with updating `worker` and `api` to `3.12` or should that be quick and straightforward?

giovanni-guidini:
> lgtm
> 
> just to make sure, this will mean we can't update the version of shared that we use to a version after this one, until the `worker` and `api` are both updated to use `3.12`?
> 
> Do we think there will be issues with updating `worker` and `api` to `3.12` or should that be quick and straightforward?

I hope it's quick and straightforward, but I'll not merge this until I do that move in case it isn't 😅

michelletran-codecov (approved):

PR #128, opened 2024-02-19
feat: add initial models + migrations for feature flags

daniel-codecov:
Update: changes in this pr have been included in [here](https://github.com/codecov/shared/pull/130) instead. 

Adds the necessary django models to implement feature flags as specified [here](https://www.notion.so/sentry/Controlled-Rollouts-2-Back-with-Experiments-1158518f09f246db9b6c0e68f8edea6c). These models essentially encapsulate what would've been hardcoded to instances of [Feature](https://github.com/codecov/shared/blob/main/shared/rollouts/__init__.py) within the code. With these models, we can now fetch/modify these values at run-time so that we can toggle on and off features without having to redeploy a new server. 

This is what the schema looks like:
Screenshot 2024-02-19 at 6 57 08 PM

Notes:
- I added indices to the FeatureFlagVariant, FeatureFlagOwnerOverride, and FeatureFlagRepoOverride tables (indicated by magnifying glass icon on fields)
- FeatureFlagOwnerOverride and FeatureFlagRepoOverride have fields `repo_ids` and `owner_ids` that are weak unenforced foreign keys to the repo and owner tables in current db. They are also of type `ArrayField` in django, not json in the diagram

What's Next:
- Need to hook this app up with `codecov-api` and `worker`, and update the `Feature` util to start using these values in the database. After hooking the models up with Django admin aswell, we could then start instrumenting features with our new feature flags and dynamically turn them on and off without redeploying a new server.

# Update:
Screenshot 2024-02-21 at 12 28 29 PM
Removed unnecessary models. Should be pretty straightforward now


shared/django_apps/rollouts/models.py
```diff
@@ -0,0 +1,50 @@
...
+    name = models.CharField(max_length=200, primary_key=True)
+    proportion = models.DecimalField(default=0, decimal_places=3, max_digits=3)
+    salt = models.CharField(max_length=10000, default=default_random_salt)
+
+    class Meta:
+        db_table = "feature_flags"
+        constraints = [
+            models.UniqueConstraint(fields=["name"], name="feature_flag_name")
+        ]
+        # TODO: assert that variant proportions sum to 1
```
daniel-codecov, line 25: not sure if this is possible

shared/django_apps/rollouts/models.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # Weak foreign keys to Owner and Respository models respectively
+    override_owner_ids = fields.ArrayField(
+        base_field=models.IntegerField(), default=list
+    )
+    override_repo_ids = fields.ArrayField(
+        base_field=models.IntegerField(), default=list
+    )
+
+    # TODO: maybe add more fields for more granularity on feature variants. EG: featureA uses value 10 vs featureB uses value 100
```
daniel-codecov, line 44: idk if this would be useful, for now probably not

shared/django_apps/rollouts/models.py
```diff
@@ -0,0 +1,50 @@
...
+        # TODO: assert that variant proportions sum to 1
+
+
+class FeatureFlagVariant(models.Model):
+    name = models.CharField(max_length=200, primary_key=True)
+    feature_flag = models.ForeignKey(
+        "FeatureFlag", on_delete=models.CASCADE, related_name="variants"
+    )
+    proportion = models.DecimalField(default=0, decimal_places=3, max_digits=3)
+    enabled = models.BooleanField(default=False)
```
daniel-codecov, line 34: I find a boolean is slightly awkward for this. What happens when we have a feature with three different variants?

shared/django_apps/rollouts/migrations/0001_initial.py
```diff
@@ -0,0 +1,155 @@
+# Generated by Django 4.2.7 on 2024-02-19 23:18
+
+import django.contrib.postgres.fields
+import django.db.models.deletion
+from django.db import migrations, models
+
+import shared.django_apps.rollouts.models
+
+
+class Migration(migrations.Migration):
```
adrian-codecov, line 10: Could you consolidate these 3 migrations into 1 migration for cleanliness? You'd basically need to 

1) delete these individual migration files
2) delete the migrations from your db entries - the `django_migrations` db table
3) remake this migration with `python manage.py makemigrations` - you can specify the app here if you want

Could you also mimic this behavior of adding the SQL that this migration would output? Sample [here](https://github.com/codecov/codecov-api/blob/5111f78d329225b7b9189feab78d4dd65335754f/codecov_auth/migrations/0051_user_customer_intent.py#L11), you can run that by doing `python manage.py sqlmigrate <app_name> <app_number>` , so for you i'd be `python manage.py sqlmigrate rollouts 0001`

Lmk if you have any questions 👌

shared/django_apps/rollouts/models.py
adrian-codecov: What are other variants you could think `enabled` could have? Is there a difference bw `enabled` and `state` or something like that?

shared/django_apps/rollouts/migrations/0001_initial.py
daniel-codecov: yeah I could combine them into one.

And yeah I can add the raw SQL to the migration as comment. Thanks

shared/django_apps/rollouts/models.py
daniel-codecov: This is a bit of a contrived example but say we had a feature like: 

`NewRateLimiter` and we had three behaviours that leaked drops from a bucket at speeds `A`, `B`, `C`. We wouldn't be able to encode those three options with a single boolean. 

Especially with the current interface which expects a boolean for: is this feature on or off pretty much
```
if LIST_REPOS_GENERATOR_BY_OWNER_SLUG.check_value(user_id, default=False):
    new_behavior()
else:
    old_behavior()
```

I think this case of having more than 2 variants is rather rare though, so its probably fine to leave it as is?

shared/django_apps/rollouts/migrations/0001_initial.py
```diff
@@ -0,0 +1,127 @@
...
+                ),
+                (
+                    "proportion",
+                    models.DecimalField(decimal_places=3, default=0, max_digits=4),
+                ),
+                (
+                    "salt",
+                    models.CharField(
+                        default=shared.django_apps.rollouts.models.default_random_salt,
+                        max_length=10000,
```
matt-codecov, line 64: the salt only need to be a few characters to effectively re-randomize the feature variant buckets. you can set this to like 32

shared/django_apps/rollouts/migrations/0001_initial.py
```diff
@@ -0,0 +1,127 @@
...
+            options={
+                "db_table": "feature_flag_variants",
+            },
+        ),
+        migrations.AddConstraint(
+            model_name="featureflag",
+            constraint=models.UniqueConstraint(
+                fields=("name",), name="feature_flag_name"
+            ),
+        ),
```
matt-codecov, line 114: this column is designated the primary key, right? doesn't that already require uniqueness?

shared/django_apps/rollouts/models.py
matt-codecov: no big deal. i think if the variant proportions only add up to, say, 0.7, all that will happen is the remaining 0.3 will get the default value. nothing will blow up

shared/django_apps/rollouts/models.py
matt-codecov: i was originally thinking the feature variant's actual value (`True`/`False`, or `A`/`B`/`C`) would still live in code, but that is also a little awkward

if we want to put the variant's value in the database, we could use a `JSONField`. if somebody enters `true` in the django admin UI it'll automatically be parsed into an actual boolean `True` in Python, or they could enter strings, arrays, objects, whatever

shared/django_apps/rollouts/models.py
matt-codecov: i don't think we need to reuse a FeatureVariant for multiple Features. the values and overrides are all feature-specific so we may as well make the variants feature-specific as well

matt-codecov (commented):
awesome work!

shared/django_apps/rollouts/models.py
```diff
@@ -0,0 +1,56 @@
...
+    proportion = models.DecimalField(default=0, decimal_places=3, max_digits=4)
+    value = models.JSONField(default=False)
+
+    # Weak foreign keys to Owner and Respository models respectively
+    override_owner_ids = fields.ArrayField(
+        base_field=models.IntegerField(), default=list, blank=True
+    )
+    override_repo_ids = fields.ArrayField(
+        base_field=models.IntegerField(), default=list, blank=True
+    )
```
daniel-codecov, line 52: The current `Feature` util uses slugs instead of ids, but it states ids [here](https://www.notion.so/sentry/Controlled-Rollouts-2-Back-with-Experiments-1158518f09f246db9b6c0e68f8edea6c)

daniel-codecov: Could just make the field a string instead of integer so that we can use both ids or slugs

PR #129, opened 2024-02-21
Bundle analysis: auto open and close DB session

JerrySentry:
Add a context manager for opening and closing DB sessions around queries in `AssetReport` and `BundleReport`. This way when the session in `BundleAnalysisReport` is closed, it won't make its children reports' session closed. This is needed in API when `BundleAnalysisReport` is resolved and cleaned up (ie DB session closed), but subsequent resolves to `AssetReport` and `BundleReport` will fail due to the shared DB session. This makes sure each `*Report` can manage its own DB session.

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/report.py
+++ b/shared/bundle_analysis/report.py
@@ -13,28 +13,14 @@ from shared.bundle_analysis import models
 from shared.bundle_analysis.parser import Parser
 
 
-def db_query(f):
-    def wrapper(*args, **kw):
-        if not args[0].db_session:
-            raise Exception(f"No DB session found for {f}")
-
-        engine = args[0].db_session.get_bind()
-        Session = sessionmaker(engine)
-        with Session() as session:
-            args[0].db_session = session
-            return f(*args, **kw)
-
-    return wrapper
-
-
 class AssetReport:
     """
     Report wrapper around a single asset (many of which can exist in a single bundle).
     """
 
-    def __init__(self, asset: models.Asset):
+    def __init__(self, db_path: str, asset: models.Asset):
+        self.db_path = db_path
         self.asset = asset
-        self.db_session = SQLAlchemySession.object_session(self.asset)
 
     @property
     def name(self):
@@ -48,15 +34,15 @@ class AssetReport:
     def size(self):
         return self.asset.size
 
-    @db_query
     def modules(self):
-        return (
-            self.db_session.query(models.Module)
-            .join(models.Module.chunks)
-            .join(models.Chunk.assets)
-            .filter(models.Asset.id == self.asset.id)
-            .all()
-        )
+        with models.get_db_session(self.db_path) as session:
+            return (
+                session.query(models.Module)
+                .join(models.Module.chunks)
+                .join(models.Chunk.assets)
+                .filter(models.Asset.id == self.asset.id)
+                .all()
+            )
 
 
 class BundleReport:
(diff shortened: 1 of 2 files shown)


shared/bundle_analysis/report.py
```diff
@@ -92,7 +95,7 @@ class BundleAnalysisReport:
     def __init__(self, db_path: Optional[str] = None):
         self.db_path = db_path
         if self.db_path is None:
-            _, self.db_path = tempfile.mkstemp()
+            _, self.db_path = tempfile.mkstemp(prefix="bundle_analysis_")
```
JerrySentry, line 98: make the temp file created for BA more identifiable. Allows API to clean of the files easier.

ajay-sentry:
I'm not too familiar with this part of the codebase, but could we have modified the __del__ method to close the session?

shared/bundle_analysis/report.py
```diff
@@ -34,54 +33,58 @@ def size(self):
...
 
     def modules(self):
-        return (
-            self.db_session.query(models.Module)
-            .join(models.Module.chunks)
-            .join(models.Chunk.assets)
-            .filter(models.Asset.id == self.asset.id)
-            .all()
-        )
+        with models.get_db_session(self.db_path) as session:
```
adrian-codecov, line 36: Will all of these open the connection, do the logic, then close the connection?

JerrySentry: Yes, the SQLAlchemy session object can act as a context manager.

shared/bundle_analysis/report.py
```diff
@@ -146,15 +149,15 @@ def metadata(self) -> Dict[models.MetadataKey, Any]:
 
     def bundle_reports(self) -> Iterator[BundleReport]:
         bundles = self.db_session.query(models.Bundle).all()
-        return (BundleReport(bundle) for bundle in bundles)
+        return (BundleReport(self.db_path, bundle) for bundle in bundles)
```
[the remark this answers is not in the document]
JerrySentry: Probably keyword is more verbose, but the variable name and the keyword is the same so I don't feel like it would be confusing to understand.

JerrySentry:
> I'm not too familiar with this part of the codebase, but could we have modified the dunder del method for this class to close the session?

@ajay-sentry do you mean have the `AssetReport` and `BundleReport` create a DB session in `__init__` and close the session in `__del__`? Yeah that should work, though the indeterministic way of knowing when the garbage collection happens have me thinking that it's not ideal. I'm worried if there's a situation where there's no connections left before previous ones gets released.

shared/bundle_analysis/report.py
```diff
@@ -34,54 +33,58 @@ def size(self):
...
     def asset_reports(self) -> Iterator[AssetReport]:
-        assets = (
-            self.db_session.query(models.Asset)
-            .join(models.Asset.session)
-            .join(models.Session.bundle)
-            .filter(models.Bundle.id == self.bundle.id)
-            .all()
-        )
-        return (AssetReport(asset) for asset in assets)
+        with models.get_db_session(self.db_path) as session:
```
ajay-sentry, line 60: > > I'm not too familiar with this part of the codebase, but could we have modified the dunder del method for this class to close the session?
> 
> @ajay-sentry do you mean have the `AssetReport` and `BundleReport` create a DB session in `__init__` and close the session in `__del__`? Yeah that should work, though the indeterministic way of knowing when the garbage collection happens have me thinking that it's not ideal. I'm worried if there's a situation where there's no connections left before previous ones gets released.

@JerrySentry yep exactly what I meant. That's a good point, probably better to have them "ad-hoc" like you have here. Though, what happens now if the connections are full when we call this line?

JerrySentry: I think it would throw an error, not exactly sure. At least what happened before when never doing a close session was that after a few sessions were opened it would spit out an error and give up.

ajay-sentry: If there's a way for us to test that scenario and make sure nothing catastrophic happens I think this lgtm :)

JerrySentry: Best way I could think of is to add a sleep in the `with`, then I made a bunch of requests at the same time. I didn't actually see any errors raised, what happened was the requests queued up till one of them was done. This is also not a true DB like the postgres we have, but it is just a mini SQLite file that is used for storing bundle stats of each commit; there shouldn't be anything catastrophic happening.

ajay-sentry (approved):
lgtm

PR #130, opened 2024-02-21
feat: use db models for feature util instead of hardcoded params [WIP]

daniel-codecov:
Moved changes from this [pr](https://github.com/codecov/shared/pull/128) to here so that new feature changes are unified.

This PR creates all the necessary changes to:
- create the new FeatureFlag and FeatureFlagVariant db models (migrations & data migrations for current Features being used)
- modify `Feature` utility to correctly fetch from the database to `check_value()` instead of using previously hardcoded parameters
- Updated most of the doc comments for `Feature` util

Tests will fail due to shared using sqlite instead of postgres


shared/rollouts/__init__.py
```diff
@@ -106,45 +114,83 @@ def _buckets(self):
...
+        Checks if the FeatureFlagVariant entries were given invalid values, which is very
+        possible since these values can be modified via Django Admin. Due to the TTL cache,
+        the values within this class will only be updated once every 5 minutes.
+        """
+        return sum(map(lambda x: x.proportion, self.ff_variants)) == 1.0 and not (
+            True
+            in map(lambda x: x.proportion < 0 or x.proportion > 1, self.ff_variants)
+        )
+
+    # NOTE: `@ttl_cache` on instance methods is ordinarily a bad idea:
```
daniel-codecov, line 158: I don't actually know if the entire spiel applies to `ttl_cache` instead of `lru_cache`, but it probably does

shared/rollouts/__init__.py
```diff
@@ -106,45 +114,83 @@ def _buckets(self):
...
     # In this case, we are okay with sharing a cache across instances, and the
     # instances are all global constants so they won't be torn down anyway.
-    @lru_cache(maxsize=50)
-    def check_value(self, identifier, default=None):
-        identifer = str(identifier)  # just in case
-        if identifier in self.overrides:
-            override_variant = self.overrides[identifier]
-            return self.variants[override_variant].value
-        elif self.proportion == 1.0 and len(self.variants) == 1:
+    @ttl_cache(maxsize=64, ttl=300)  # 5 minute ttl
```
matt-codecov, line 166: hmm. i think this will wind up making more database queries than we want. consider the following:
- we call `feature.check_value(A)`. we'll fetch the model from the db, compute which variant `A` belongs in, put it in the cache, and then return
- we immediately call `feature.check_value(B)`. `B` isn't in our cache, so we run the function again. we make a db request, compute the variant, cache it, and then return

what i think we want instead is to wrap accesses to the model in a `get_model()` function or something and put that behind a second cache, and the second cache has a TTL. something like:
```
# this caches the variant that each identifier should use, assuming the db
# model is up to date
@lru_cache(maxsize=64)
def check_value(self, identifier, default=None):
    model = self.model
    override_variant = self.get_override_variant(model)
    if override_variant:
        ... # and so on

# this will re-fetch the db model if it's stale. if anything in the model changed,
# we should clear the check_value() cache
#
# the @property means it's accessed as if it's a member variable: self.model
@property
@ttl_cache(maxsize=64, ttl=300)
def model(self):
    new_model = self.fetch_from_db()
    if new_model != self._model:
        self._model = new_model
        self.check_value.cache_clear()
    return self._model
```

shared/rollouts/__init__.py
```diff
@@ -106,45 +114,83 @@ def _buckets(self):
...
+        """
+        for variant in self.ff_variants:
+            if (
+                identifier in variant.override_owner_ids
+                or identifier in variant.override_repo_ids
+            ):
+                return variant
+        return None
+
+    def _is_valid_rollout(self):
```
matt-codecov, line 147: maybe we run this in the function that fetches values from the DB, and if it isn't valid log a warning. that way somebody making a mistake in the web UI (or the cache clearing while they're still in the middle of making updates) doesn't totally shut down every feature

we use the python standard library logger a la https://github.com/codecov/worker/blob/main/helpers/checkpoint_logger/__init__.py#L22

shared/rollouts/__init__.py
```diff
@@ -80,23 +70,41 @@ class Feature:
...
+
+            self.feature_flag = FeatureFlag(**args)
+            self.feature_flag.save()
+        else:
+            # ignore constructor args if db entry already exists.
+            self.feature_flag = ff_query.first()
+
+        self.ff_variants = list(
+            FeatureFlagVariant.objects.filter(feature_flag=self.feature_flag.name)
+        )
```
matt-codecov, line 95: you could probably just make this the definition of `_fetch_and_set_from_db()` and reuse that function here

shared/rollouts/__init__.py
matt-codecov: i think it does!

matt-codecov (commented):
very solid, thanks for updating the doc comments too! i think it needs to be reworked a bit before merging but this is good progress

once this is ready to merge, there's some non-coding work we might want to do. there are three usages of `Feature` already sort-of in use: [two in worker](https://github.com/codecov/worker/blob/f0c512d48558ef6d9fb8b7fea1dec1f9e1ea431c/rollouts/__init__.py#L16-L35) and [one in api](https://github.com/codecov/codecov-api/blob/64a6de2beb0e14ce2fc0d13ffb12cec47fc45dab/utils/rollouts.py#L10). if we just update the code and remove their variant definitions, those features will be totally disabled until we can hurry into the django admin UI and recreate them. we should try to avoid that if we can. probably the cleanest way is by creating [an additional migration to create the db rows for those features/variants](https://docs.djangoproject.com/en/1.8/topics/migrations/#data-migrations)

shared/rollouts/__init__.py
daniel-codecov: ah I see my mistake, thanks!

shared/django_apps/rollouts/migrations/0002_auto_20240223_1522.py
```diff
@@ -0,0 +1,42 @@
...
+    FeatureFlagVariant.objects.create(
+        name="use_label_index_in_report_processing_enabled",
+        feature_flag=use_label_index_in_report_processing,
+        proportion=1.0,
+        value=True,
+        override_repo_ids=[
+            "github/giovanni-guidini/sentry",
+            "github/giovanni-guidini/components-demo",
+        ],
+    )
```
daniel-codecov, line 33: to seed for these features:

https://github.com/codecov/worker/blob/f0c512d48558ef6d9fb8b7fea1dec1f9e1ea431c/rollouts/__init__.py#L16-L35

shared/django_apps/rollouts/migrations/0002_auto_20240223_1522.py
```diff
@@ -0,0 +1,42 @@
...
+    FeatureFlagVariant.objects.create(
+        name="use_label_index_in_report_processing_enabled",
+        feature_flag=use_label_index_in_report_processing,
+        proportion=1.0,
+        value=True,
+        override_repo_keys=[
+            "github/giovanni-guidini/sentry",
+            "github/giovanni-guidini/components-demo",
+        ],
+    )
```
daniel-codecov, line 33: to seed these features: https://github.com/codecov/worker/blob/f0c512d48558ef6d9fb8b7fea1dec1f9e1ea431c/rollouts/__init__.py#L16-L35

matt-codecov (approved):

PR #136, opened 2024-02-27
Daniel/postgres docker setup

daniel-codecov:
Dockerizes shared so that we can spin up a `postgres` and `timeseriesdb` container to run our tests that use postgres-specific features. Updates our github actions so that they use docker compose instead of local installations. 

Notes:
- not sure where/how we can cache some docker layers/images


.github/workflows/ci.yml
```diff
@@ -62,45 +61,23 @@ jobs:
...
-      - name: Install reqs
+      - name: Install docker compose
         run: |
-          . venv/bin/activate
-          pip install setuptools_rust
-          pip install -r tests/requirements.txt
-          pip install -r requirements.txt
-          python setup.py develop
-          pip install codecov-cli
+          sudo curl -SL https://github.com/docker/compose/releases/download/v2.20.0/docker-compose-linux-x86_64 -o /usr/local/bin/docker-compose
```
adrian-codecov, line 73: Is it normal to run sudo on ci workflows?

matt-codecov (approved):

PR #139, opened 2024-03-01
... :O Port Some API Models to Shared

adrian-codecov:
TLDR, we want to move all django API models to shared so they're shared amongst different codebases. This is a stab at it, first taking a look at `Core`, `Codecov_auth`and `Reports` apps. 

Specifically,
- I've moved all models inside those apps
-   There are some fields I didn't move as those would require a substantially higher effort. I'm abstaining from that and following [this](https://www.notion.so/sentry/Django-Model-Migration-to-Shared-795b38364fd04253a21d52bebe7f7b8a?pvs=4#7acced0486af4b299a731b84fad94147) approach instead
- I moved all migrations for all those apps
-   All the `makemigrations` commands had noop
- Moved some helper files that made sense for this transition. Didn't explicitly test those fns as they're tested through other functions in API

### Files to take a special look at
- docker-compose.yml
- setup.py
- duimmy_settings.py
- migrate.py (mostly copy paste but slight change from original file)

The rest of the files are mostly copy-paste models + migrations

### Note
I had to adjust some imports in migrations independently, for example, they used to be `from core.models import DateTimeWithouTZField` to from shared.django_apps.core.models import DateTimeWithouTZField`

### Legal Boilerplate


docker-compose.yml
```diff
@@ -31,6 +33,10 @@ services:
         tmpfs:
           size: 1024M
 
+  redis:
```
adrian-codecov, line 36: Needed to run migrations tests

shared/django_apps/dummy_settings.py
```diff
@@ -8,18 +8,63 @@
...
+    "shared.django_apps.legacy_migrations",
     "shared.django_apps.pg_telemetry",
     "shared.django_apps.ts_telemetry",
     "shared.django_apps.rollouts",
+    # API models
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.messages",
+    "django.contrib.postgres",
```
matt-codecov, line 20: why were these necessary again? weren't they looped in because of some helper function that used a template or something? can we get rid of them?

shared/django_apps/dummy_settings.py
```diff
@@ -8,18 +8,63 @@
...
+    "shared.django_apps.codecov_auth",
+    "shared.django_apps.core",
+    "shared.django_apps.reports",
 ]
 
-MIDDLEWARE = []
+# Migrated from API to get Minio working
+MINIO_ACCESS_KEY = get_config("services", "minio", "access_key_id")
+MINIO_SECRET_KEY = get_config("services", "minio", "secret_access_key")
+MINIO_LOCATION = "codecov.s3.amazonaws.com"
```
matt-codecov, line 29: i don't think this is ever referenced. and we don't use aws at all, do we? so i think this might be dead code

shared/django_apps/dummy_settings.py
```diff
@@ -8,18 +8,63 @@
...
+MINIO_HASH_KEY = get_config("services", "minio", "hash_key")
+ARCHIVE_BUCKET_NAME = "codecov"
+
+MIDDLEWARE = [
+    "django.contrib.auth.middleware.AuthenticationMiddleware",
+    "django.middleware.common.CommonMiddleware",
+    "django.middleware.csrf.CsrfViewMiddleware",
+    "django.contrib.messages.middleware.MessageMiddleware",
+    "django.contrib.sessions.middleware.SessionMiddleware",
```
matt-codecov, line 38: are any of these still necessary?

shared/django_apps/dummy_settings.py
```diff
@@ -8,18 +8,63 @@
...
+        "APP_DIRS": True,
+        "OPTIONS": {
+            "context_processors": [
+                "django.template.context_processors.debug",
+                "django.template.context_processors.request",
+                "django.contrib.auth.context_processors.auth",
+                "django.contrib.messages.context_processors.messages",
+            ]
+        },
+    }
```
matt-codecov, line 55: is this still necessary?

codecov.yml
```diff
@@ -11,8 +11,20 @@ ignore:
...
+    hash_key: 12345bf3f7d947f2a0681b2154067890
+    verify_ssl: false
+    host: "minio"
+    port: 9000
+    # bucket: <bucket-name>
+    # region: <bucket-region>
+    access_key_id: codecov-default-key
+    secret_access_key: codecov-default-secret
+    client_uploads: true
+
```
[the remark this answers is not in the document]
matt-codecov: i know there is some mock config stuff set up in https://github.com/codecov/shared/blob/main/tests/conftest.py but i don't actually know if/where there is a yml for the docker-compose setup. it may be that nobody needed one. if we need one now, put it in the `docker` folder?

matt-codecov (commented):
looks like there are a lot of totally untested files in this PR. could you check if any of them have tests in API you can steal?
- `api_archive/archive.py`
- `api_archive/storage.py`
- `django_apps/codecov_auth/helpers.py`
- `django_apps/codecov_auth/managers.py`
- `django_apps/core/managers.py`
- anything in `django_apps/utils`
- `reports/api_report_service.py`

doesn't necessarily have to be in this PR, but would be good not to lose coverage. i guess when some of this code is deleted from API we'll be able to find tests that cover them

also i think there is a ton of code here we can dedup beyond models lol

setup.py
```diff
@@ -50,5 +50,13 @@
...
         "cachetools",
         "django-better-admin-arrayfield",
+        # API Deps
+        "django-prometheus",
+        "django-model-utils",
+        "factory-boy",
+        "python-redis-lock",
+        "django-model-utils==4.3.1",
+        "requests==2.31.0",
+        "sentry-sdk>=1.40.0",
```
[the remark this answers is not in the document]
adrian-codecov: Only found factory-boy to be test, the others are used during runtime

shared/api_archive/archive.py
```diff
@@ -0,0 +1,268 @@
...
+from minio import Minio
+
+from shared.api_archive.storage import StorageService
+from shared.config import get_config
+from shared.utils.ReportEncoder import ReportEncoder
+
+log = logging.getLogger(__name__)
+
+
+class MinioEndpoints(Enum):
```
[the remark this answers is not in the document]
adrian-codecov: It probably does lol, I'll leave a note here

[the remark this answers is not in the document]
adrian-codecov: I'm keeping track of TODOs here, https://github.com/codecov/engineering-team/issues/1516

shared/django_apps/dummy_settings.py
adrian-codecov: admin -> `LogEntry` used in codecov_auth helper
contentypes -> get_gitlab_url used in codecov_auth helper
postgres -> I get test errors otherwise, `syntax error at or near "gin_trgm_ops"`, https://stackoverflow.com/questions/37334902/postgres-syntax-error-creating-index-using-prgm-gin

The rest I've deleted and I'll test tmrw

shared/django_apps/dummy_settings.py
adrian-codecov: Deleted and ran test, no errors. I'll rebuild API with this image tomorrow and test

shared/django_apps/dummy_settings.py
adrian-codecov: Deleted and ran test, no errors. I'll rebuild API with this image tomorrow and test

shared/django_apps/dummy_settings.py
adrian-codecov: Deleted and ran test, no errors. I'll rebuild API with this image tomorrow and test

codecov.yml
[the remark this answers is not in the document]
adrian-codecov: `hash_key` is fake, I manually used prepend/append 12345 67890 to make it fake

Its late so not fully following the rest of the ask here, should I move this to a docker folder?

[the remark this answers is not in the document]
matt-codecov: i think this `codecov.yml` is the one that defines our flags and stuff for our CI uploads, not the `codecov.yml` that defines the service config inside docker-compose. i don't think this will be read by anything. if you just delete it, does everything still work?

matt-codecov (approved):

PR #140, opened 2024-03-01
feat: add helper text

daniel-codecov:
Adds helper text to `proportion` and`value`.
 
Screenshot 2024-03-01 at 11 31 32 AM

Closes the first three here: https://github.com/codecov/engineering-team/issues/1315

### Legal Boilerplate


shared/django_apps/dummy_settings.py
```diff
@@ -2,7 +2,7 @@
 
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
 BASE_DIR = Path(__file__).resolve().parent
-from .db_settings import *
+from shared.django_apps.db_settings import *
```
daniel-codecov, line 5: it doesn't like the relative import

shared/django_apps/rollouts/migrations/0003_alter_featureflag_proportion_and_more.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 4.2.7 on 2024-03-01 16:29
```
daniel-codecov, line 1: doesn't actually alter db, just adds helper text for django

shared/django_apps/dummy_settings.py
```diff
@@ -27,15 +27,15 @@
...
+        "NAME": "postgres",
         "USER": "postgres",
         "PASSWORD": "password",
         "HOST": "postgres",
         "PORT": 5432,
     },
     "timeseries": {
         "ENGINE": "django.db.backends.postgresql",
-        "NAME": "POSTGRES_USER",
+        "NAME": "postgres",
```
daniel-codecov, line 38: this was forgotten in https://github.com/codecov/shared/commit/9c1a9da61a778b1e04c89503df501a507c7339c2

joseph-sentry (approved):
lgtm

PR #153, opened 2024-03-13
Bundle Analysis: surface plugin name on parser error

JerrySentry:
Inject the plugin name (the plugin that was responsible for creating the stats file) if parsing the bundle stats file fails, this will be later used by the worker's BA processor to surface the bad plugin name as Sentry metrics.

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/parser.py
+++ b/shared/bundle_analysis/parser.py
@@ -68,7 +68,8 @@ class Parser:
 
             with open(path, "rb") as f:
                 for event in ijson.parse(f):
-                    self._parse_event(event)
+                    print("PARSING EVENT")
+                    self._parse_item(event)
 
                 # Delete old session/asset/chunk/module with the same bundle name if applicable
                 old_session = (
@@ -127,7 +128,7 @@ class Parser:
         elif prefix == "duration":
             self.info["duration"] = value
 
-    def _parse_event(self, event: Tuple[str, str, str]):
+    def _parse_item(self, event: Tuple[str, str, str]):
         prefix, _, value = event
         prefix_path = prefix.split(".")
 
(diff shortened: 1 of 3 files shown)


shared/bundle_analysis/parser.py
```diff
@@ -54,66 +54,57 @@ def reset(self):
...
+                self.session.info = json.dumps(self.info)
+
+                # this happens last so that we could potentially handle any ordering
+                # of top-level keys inside the JSON (i.e. we couldn't associate a chunk
+                # to an asset above if we parse the chunk before the asset)
+                self._create_associations()
+
+                assert self.session.bundle is not None
+                return self.session.id
+        except Exception as e:
```
JerrySentry, line 96: The diff is just formatting, basically wrapped the entire code block in a `try/except`, inject a plugin name to whatever the exception raised during bundle file parsing is.

shared/bundle_analysis/parser.py
```diff
@@ -54,66 +54,57 @@ def reset(self):
...
+            # Inject the plugin name to the Exception object so we have visibilitity on which plugin
+            # is causing the trouble.
+            e.bundle_analysis_plugin_name = self.info.get("plugin_name", "unknown")
+            raise e
 
     def _parse_event(self, event: Tuple[str, str, str]):
         prefix, _, value = event
```
JerrySentry, line 103: Shifted the ordering of `if` statements for readability, it follows the same sequence of what is on the bundle stats file.

adrian-codecov: Would altering the order change how this is decided? Is it relevant to check assets/chunks/modules before we check the version,, name, duration, etc?

adrian-codecov (approved):

shared/bundle_analysis/parser.py
JerrySentry: We want to grab the info asap because if/when the parsing for ACM fails then we know which bundler plugin failed via the info.bundler_name we collected already

PR #157, opened 2024-03-14
upper limit on django version

nora-codecov:
Without an upper limit here on django versioning, it goes to latest release, `5.x`, which is incompatible with python `3.9`, the version `api` uses. Out of an abundance of caution, let's keep these two projects on the same major version, which has the security patch for the [target vuln](https://github.com/advisories/GHSA-xxj9-f6rv-m3x4)


JerrySentry (approved):
that is good to know

PR #158, opened 2024-03-14
chore: update async interface

daniel-codecov:
Updates the async version of `check_value()` so that it matches the non-async version.

### DIFF
--- a/shared/rollouts/__init__.py
+++ b/shared/rollouts/__init__.py
@@ -187,7 +187,7 @@ class Feature:
             )
         )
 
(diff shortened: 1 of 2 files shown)


shared/bundle_analysis/models.py
```diff
@@ -98,6 +97,8 @@
 For now if the SQLAlchemy version is <1.4 it will go through the custom LegacySessionManager
 context manager object to handle opening and closing its sessions.
 """
+
```
daniel-codecov, line 100: lint

matt-codecov (approved):

PR #161, opened 2024-03-22
fix: fix glob pattern translation

giovanni-guidini:
Glob patterns were always considered as "prefixes",
and would match any depth after their ending.

Recently this became an issue noticed by a customer that was
trying to ignore all files ending with `bundle`, but ended up having
more files than intended ignored.

These changes remove this "glob is always a prefix" idea and solve this
use case. Users can still use "path as a prefix" but only proper paths,
not globs. They can also use regexes.

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/models.py
+++ b/shared/bundle_analysis/models.py
@@ -86,7 +86,6 @@ SCHEMA_VERSION = 1
 Base = declarative_base()
 
 
-
 """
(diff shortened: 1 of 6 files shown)


joseph-sentry (approved):
I get the feeling this may break many customers that have been relying on this incorrect behaviour, this change should definitely come with a change in the docs that calls this out as something that was done recently

PR #162, opened 2024-03-25
chore: add repo logs when fetching GH repos

ajay-sentry:
Adding some logging for the fetch_page functions in github.py for logging the repo names when returned from the api call

### Legal Boilerplate

### DIFF
--- a/shared/torngit/github.py
+++ b/shared/torngit/github.py
@@ -645,7 +645,9 @@ class Github(TorngitBaseAdapter):
 
         log.info(
             "Fetched page of repos using installation",
-            extra=dict(page_size=page_size, page=page, repos=repos),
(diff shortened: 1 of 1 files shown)


shared/torngit/github.py
```diff
@@ -732,9 +751,9 @@ async def get_repos_from_nodeids_generator(
                             primary_language.get("name") if primary_language else None
                         ),
                         "private": raw_repo_data["isPrivate"],
-                        "branch": default_branch.get("name")
-                        if default_branch
-                        else None,
+                        "branch": (
```
ajay-sentry, line 754: The stuff here and below is just formatting stuff from VSCode

adrian-codecov (approved):

PR #164, opened 2024-03-27
feat: parallel upload processing

daniel-codecov:
Not much, most of the work is done [here](https://github.com/codecov/worker/pull/347). Also a small quality of life change for refreshing feature flags when developing.

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/models.py
+++ b/shared/bundle_analysis/models.py
@@ -85,7 +85,6 @@ SCHEMA_VERSION = 1
 
 Base = declarative_base()
 
-
 """
 Create a custom context manager for SQLAlchemy session because worker is currently
 stuck on SQLAlchemy version <1.4, and built in context manager for session is introduced
--- a/shared/reports/enums.py
+++ b/shared/reports/enums.py
@@ -7,7 +7,9 @@ class UploadState(CodecovDatabaseEnum):
     ERROR = (3,)
     FULLY_OVERWRITTEN = (4,)
     PARTIALLY_OVERWRITTEN = (5,)
-    PARALLEL_PROCESSED = (6,)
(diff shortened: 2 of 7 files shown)


shared/reports/resources.py
```diff
@@ -1021,18 +1021,29 @@ def manifest(self):
...
-        sessionid = self.next_session_number()
+    def add_session(self, session, parallel_idx=None):
+        sessionid = self.next_session_number(parallel_idx)
         self.sessions[sessionid] = session
         if self._totals:
             # add session to totals
-            self._totals = dataclasses.replace(self._totals, sessions=sessionid + 1)
+            if parallel_idx:
+                self._totals = dataclasses.replace(
+                    self._totals, sessions=max(sessionid + 1, self._totals.sessions)
```
daniel-codecov, line 1042: I'm not actually sure if `max` here is necessary

shared/bundle_analysis/models.py
```diff
@@ -85,8 +85,6 @@
 
 Base = declarative_base()
 
-
```
daniel-codecov, line 88: lint

shared/reports/enums.py
```diff
@@ -7,6 +7,9 @@ class UploadState(CodecovDatabaseEnum):
     ERROR = (3,)
     FULLY_OVERWRITTEN = (4,)
     PARTIALLY_OVERWRITTEN = (5,)
+    PARALLEL_PROCESSED = (
+        6,
+    )  # not used right now, but will when parallel upload procesing is rolled out
```
matt-codecov, line 12: what is the difference between this and `PROCESSED`? could you explain when this status will be used in a comment?

matt-codecov (approved):

PR #166, opened 2024-04-02
feat: support endpoint URL in AWS storage

imnotjames:
this allows for using other s3 compatible servers with the AWS config

### Legal Boilerplate


imnotjames:
Testing with this against garage.

imnotjames:
Well, this doesn't matter anyway - this code is never used in the API.  The API will only ever use minio.

PR #167, opened 2024-04-02
feat: support minio region passing to client

imnotjames:
this allows us to specify the region correctly so various scopes and other aspects don't default to us-east-1 internally

### Legal Boilerplate

### DIFF
--- a/shared/storage/minio.py
+++ b/shared/storage/minio.py
@@ -37,6 +37,7 @@ class MinioStorageService(BaseStorageService):
             self.minio_config["verify_ssl"],
             self.minio_config.get("iam_auth", False),
             self.minio_config["iam_endpoint"],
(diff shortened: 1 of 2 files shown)


imnotjames:
Oh, region should be a get as it's optional.  I'll update that.

imnotjames:
FIxed tests - they weren't expecting region & I was missing a test to validate that region was getting passed through as expected.

imnotjames:
🥲 okay.  Thought I fixed tests.  This is what I get for relying on CI to validate my tests.  I'll get another commit up soon for that.

trent-codecov (approved):
This looks good to me. Adds the ability to configure this while preserving existing defaults.

Thanks for the contribution!

PR #169, opened 2024-04-05
fix: bundle deletion deadlock

JerrySentry:
### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/models.py
+++ b/shared/bundle_analysis/models.py
@@ -231,9 +231,7 @@ class Chunk(Base):
     initial = Column(types.Boolean, nullable=False)
 
     session = relationship("Session", backref=backref("chunks"))
-    assets = relationship(
-        "Asset", secondary=assets_chunks, back_populates="chunks", cascade="all, delete"
-    )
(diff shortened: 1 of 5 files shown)


shared/bundle_analysis/models.py
```diff
@@ -231,9 +231,7 @@ class Chunk(Base):
     initial = Column(types.Boolean, nullable=False)
 
     session = relationship("Session", backref=backref("chunks"))
-    assets = relationship(
-        "Asset", secondary=assets_chunks, back_populates="chunks", cascade="all, delete"
-    )
+    assets = relationship("Asset", secondary=assets_chunks, back_populates="chunks")
```
adrian-codecov, line 234: Is this a relationship that's already there in API?

adrian-codecov (approved):

shared/bundle_analysis/models.py
adrian-codecov: 🤦 yep yep

adrian-codecov: I'm too deep in migrations land, cannot not think about these 😂

PR #171, opened 2024-04-05
[chore] switch from `sleep 86400` to a tty service

matt-codecov:
instead of starting a container and running `sleep 86400` to keep it from shutting down, we can set `tty: true`

### Legal Boilerplate


giovanni-guidini (approved):
Today I learned.

Why is codecov considering this file as "covered" I wonder... shouldn't be in any tests report, should it?
[edit] I checked the UI and it's claiming indirect changes, not this file being changed.

PR #172, opened 2024-04-05
fix: fix feature bucketing logic

daniel-codecov:
Should close https://github.com/codecov/engineering-team/issues/1515. Refactors the bucketing logic so that users' assignments to variants never change even when the total rollout proportions change.

Added a test for this case specifically

Screenshot 2024-04-05 at 3 04 15 PM

Screenshot 2024-04-05 at 3 04 29 PM

### Legal Boilerplate

### DIFF
--- a/tests/unit/test_rollouts.py
+++ b/tests/unit/test_rollouts.py
@@ -39,7 +39,11 @@ class TestFeature(TestCase):
             # should be [0..33], [34..66], and [67..100]. Anything from [101..200]
             # is not part of the rollout yet.
             buckets = complex_feature._buckets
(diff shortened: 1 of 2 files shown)


tests/unit/test_rollouts.py
```diff
@@ -39,7 +39,11 @@ def test_buckets(self):
             # should be [0..33], [34..66], and [67..100]. Anything from [101..200]
```
ajay-sentry, line 39: nit: I think this comment needs to be updated, right?

daniel-codecov: ah good catch

matt-codecov (approved):
beautiful

PR #178, opened 2024-04-10
feat: add user measurements in shared

adrian-codecov:
This ticket is mostly to move + mildly adjust merged changes from api to shared. Namely, those changes are the ones from these 2 prs: https://github.com/codecov/codecov-api/pull/414 and https://github.com/codecov/codecov-api/pull/425.

We're moving the user_measurement + query/insert logic to this table from API to Shared, as well as the partitioning manager.

The user_measurements table exists in prod + staging but they don't have any data, so I'll coordinate with ops to delete those tables, run fresh partitions + set this model as a fresh model from fresh rather than a prexisting api model for ease.

### Legal Boilerplate


shared/django_apps/dummy_settings.py
```diff
@@ -21,10 +28,29 @@
     "shared.django_apps.reports",
 ]
 
-MIDDLEWARE = []
-
-TEMPLATES = []
+# Needed for makemigrations to work
```
adrian-codecov, line 31: Had to bring these 2 back since the `makemigrations` command complained otherwise

shared/django_apps/dummy_settings.py
```diff
@@ -68,6 +94,12 @@
 
 AUTH_PASSWORD_VALIDATORS = []
 
+# See https://django-postgres-extra.readthedocs.io/en/master/settings.html
+POSTGRES_EXTRA_DB_BACKEND_BASE: "django_prometheus.db.backends.postgresql"
```
adrian-codecov, line 98: So we didn't have the engine set to `django_prometheus` like api did here, https://github.com/codecov/codecov-api/pull/414/files#diff-82bf6dbe04b3ff913e6d14636041575bd98d8363d424e0168e72bb8ab7c5e700L201, but we want to keep the DB settings aligned don't we?

shared/django_apps/user_measurements/partitioning.py
```diff
@@ -0,0 +1,27 @@
...
+    PostgresCurrentTimePartitioningStrategy,
+    PostgresPartitioningManager,
+    PostgresTimePartitionSize,
+)
+from psqlextra.partitioning.config import PostgresPartitioningConfig
+
+from shared.django_apps.user_measurements.models import UserMeasurement
+
+# Overlapping partitions will cause errors - https://www.postgresql.org/docs/current/ddl-partitioning.html#DDL-PARTITIONING-DECLARATIVE -> "create partitions"
+manager = PostgresPartitioningManager(
```
joseph-sentry, line 12: does this specific manager object get used anywhere

shared/upload/utils.py
```diff
@@ -0,0 +1,56 @@
...
+    CLI = "cli"
+
+
+def query_monthly_coverage_measurements(plan_service: PlanService) -> int:
+    owner = plan_service.current_org
+    queryset = UserMeasurement.objects.filter(
+        owner=owner,
+        private_repo=True,
+        created_at__gte=timezone.now() - timedelta(days=30),
+        report_type="coverage",
```
joseph-sentry, line 25: maybe this can be `ReportType.COVERAGE.value`

shared/django_apps/user_measurements/partitioning.py
```diff
@@ -0,0 +1,27 @@
...
+        # 12 partitions ahead, each partition is 1 month
+        # Partitions can be deleted after 12 months of their starting date, not their creation, via the pgpartition command.
+        # They won't be automatically deleted though.
+        # Partitions will be named `[table_name]_[year]_[3-letter month name]`.
+        PostgresPartitioningConfig(
+            model=UserMeasurement,
+            strategy=PostgresCurrentTimePartitioningStrategy(
+                size=PostgresTimePartitionSize(months=1),
+                count=12,
+                max_age=relativedelta(months=12),
```
joseph-sentry, line 23: is this not insinuated by the count above?

shared/django_apps/user_measurements/models.py
```diff
@@ -0,0 +1,56 @@
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    uploader_used = models.CharField()
+    private_repo = models.BooleanField()
+    report_type = models.CharField(
+        null=True, max_length=100, choices=ReportType.choices
+    )
+
+    class Meta:
+        db_table = "user_measurements"
+        indexes = [
```
joseph-sentry, line 45: good indexes 👍

joseph-sentry (approved):
looks good, just had some questions

shared/django_apps/user_measurements/partitioning.py
adrian-codecov: Yeap, this is the code that enables running the `python manage.py pgpartition` command, which automatically creates the partitions specified in these settings. I followed the config instructions of the docs, https://django-postgres-extra.readthedocs.io/en/latest/table_partitioning.html#configuration

I also centralized the logic here in case the partition logic wants to be called in any of its child repositories

shared/upload/utils.py
adrian-codecov: yeap good callout, pushing in a bit

shared/django_apps/user_measurements/partitioning.py
adrian-codecov: Yeah this settings is mostly a "after how long are these partitions safely deletable by the partition command". So these could be 24 or any non 12 number for instance, so while they're the same number, they are distinct. This also means that the April 2024 partition is deletable in april 2025, and the may 2024 one in may 2025 and so on

PR #179, opened 2024-04-11
add github api call counter

nora-codecov:
Part of https://github.com/codecov/engineering-team/issues/1167, this is for Github calls, the other git providers will be added next.

We want to know who's making all these calls! So I added some `Counters`.

This is a first implementation - the strategy I picked was to increment the `Counter` before the code makes the request to github - it uses a helper function and technically could error out somewhere along the way before it actually makes the call. So you could argue that the `Counters` are counting _intended_ requests, but hopefully errors between intent-to-call-github and actually making the outgoing request are low, and if not, at least they are logged.

I learned a bit from the Prometheus docs - each `label` value on a `Counter` is an individual object that will live and wait around to be incremented (and take up resources). I'm going to try initializing them all in the init method of the `Github` object, I think this is the right way to use Counters ([reference](https://prometheus.github.io/client_python/instrumenting/labels/)). Additionally, the docs say a Counter shouldn't have more than 10 labels ([reference](https://prometheus.io/docs/practices/instrumentation/#use-labels)), and this one is ~50, so we will need to monitor prometheus performance around this deploy.

### DIFF
--- a/shared/torngit/base.py
+++ b/shared/torngit/base.py
@@ -50,33 +50,6 @@ class TorngitBaseAdapter(object):
         "xtend",
     )
 
-    def get_client(self, timeouts: List[int] = []) -> httpx.AsyncClient:
-        if timeouts:
-            timeout = httpx.Timeout(timeouts[1], connect=timeouts[0])
-        else:
-            timeout = httpx.Timeout(self._timeouts[1], connect=self._timeouts[0])
-        return httpx.AsyncClient(
-            verify=self.verify_ssl
-            if not isinstance(self.verify_ssl, bool)
-            else self.verify_ssl,
-            timeout=timeout,
-        )
-
-    def get_token_by_type(self, token_type: TokenType):
-        if self._token_type_mapping.get(token_type) is not None:
-            return self._token_type_mapping.get(token_type)
-        return self.token
-
-    def get_token_by_type_if_none(self, token: Optional[str], token_type: TokenType):
-        if token is not None:
-            return token
-        return self.get_token_by_type(token_type)
-
-    def _oauth_consumer_token(self):
-        if not self._oauth:
-            raise Exception("Oauth consumer token not present")
-        return self._oauth
-
     def __init__(
         self,
         oauth_consumer_token: OauthConsumerToken = None,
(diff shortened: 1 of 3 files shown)


shared/torngit/base.py
```diff
@@ -107,6 +80,33 @@ def __repr__(self):
...
+    def get_token_by_type_if_none(self, token: Optional[str], token_type: TokenType):
+        if token is not None:
+            return token
+        return self.get_token_by_type(token_type)
+
+    def _oauth_consumer_token(self):
+        if not self._oauth:
+            raise Exception("Oauth consumer token not present")
+        return self._oauth
+
```
nora-codecov, line 109: I didn't change any logic, just rearranged

shared/torngit/github.py
```diff
@@ -166,30 +317,30 @@ def get_api_url(cls):
...
         return self._token
 
-    @property
-    def api_host_header(self):
-        return self.get_api_host_header()
-
-    @property
-    def host_header(self):
-        return self.get_host_header()
-
```
nora-codecov, line 192: I didn't change any logic, just rearranged

tests/unit/torngit/test_github.py
```diff
@@ -262,14 +257,11 @@ async def test_make_http_call_change_host(
...
         assert client.request.call_count == 1
         args, kwargs = client.request.call_args
-        print(args)
-        print(kwargs)
         assert kwargs.get("headers") is not None
         assert kwargs.get("headers").get("Host") == "github.com"
         assert len(args) == 2
         built_url = args[1]
         parsed_url = urlparse(built_url)
-        print(parsed_url)
```
nora-codecov, line 272: There were several print statements in this test file, I removed them, let me know if you want me to add them back.

shared/torngit/github.py
```diff
@@ -37,6 +37,22 @@
...
+)
+
+GITHUB_API_ENDPOINTS = {
+    "test_2": GITHUB_API_CALL_COUNTER.labels(endpoint="test_2"),
+    "test_4": GITHUB_API_CALL_COUNTER.labels(endpoint="test_4"),
+}
+
+GITHUB_API_CALL_COUNTER.labels(endpoint="test_3")
+GITHUB_API_CALL_COUNTER.labels(endpoint="test_3").inc()
+GITHUB_API_ENDPOINTS["test_4"].inc()
```
adrian-codecov, line 53: What are these tests?

shared/torngit/github.py
```diff
@@ -135,22 +151,167 @@ def prepare(self, query_name: str, variables: dict) -> Optional[dict]:
 class Github(TorngitBaseAdapter):
     service = "github"
     graphql = GitHubGraphQLQueries()
-    urls = dict(
```
adrian-codecov, line 138: Took me a sec to figure out, but we're not using these altogether it seems?

adrian-codecov (commented):
Left quick q, otherwise seems good! Is this testable in staging?

shared/torngit/github.py
nora-codecov: yep! Turns out only 1 of those was in use, the other endpoints we use are written out in each function. The refactor I'm going to do is going to include something to replace this - a collection of the urls we use

PR #182, opened 2024-04-12
fix: remove parallel_processed enum

daniel-codecov:
Remove the `PARALLEL_PROCESSED` enum because it currently is not used. It will be used when parallel upload processing is fully rolled out.

There's no need to make the migration for it right now if it's not going to be used. (migration hasn't been made yet)

  Sentry/Codecov employees and contractors can delete or ignore the following.

-->

### Legal Boilerplate

### DIFF
--- a/shared/torngit/github.py
+++ b/shared/torngit/github.py
@@ -1245,7 +1245,9 @@ class Github(TorngitBaseAdapter):
                     (
                         "\ndeleted file mode 100644"
                         if f["status"] == "removed"
-                        else "\nnew file mode 100644" if f["status"] == "added" else ""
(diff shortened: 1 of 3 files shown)


shared/torngit/github.py
```diff
@@ -1245,7 +1245,9 @@ async def get_compare(
                     (
                         "\ndeleted file mode 100644"
                         if f["status"] == "removed"
-                        else "\nnew file mode 100644" if f["status"] == "added" else ""
```
daniel-codecov, line 1248: lint

joseph-sentry (approved):

PR #185, opened 2024-04-13
Add count_and_get_url_template to github handler

nora-codecov:
Part of https://github.com/codecov/engineering-team/issues/1167, this is for Github calls, the other git providers will be added next.

We want to know who's making all these calls! So I added some Counters.

This is a first implementation - the strategy I picked was to increment the Counter before the code makes the request to github - it uses a helper function and technically could error out somewhere along the way before it actually makes the call. So you could argue that the Counters are counting intended requests, but hopefully errors between intent-to-call-github and actually making the outgoing request are low, and if not, at least they are logged.

The counts are named for the functions that call them, so we can trace a`Counter` back to the place in our code where it is being called.

An additional task got pulled into this pr - organize and track the github urls we are using, which made the lines in this pr balloon (sorry!). Writing the `Counter` code, I realized that the `urls` dict on the `Github` object was not being used - each function that called an endpoint had a url written as a str. I removed those, put them in a collection, and made them into str `Templates` - now every url has a counter and a template, the counter is incremented when the url is built.

I learned a bit from the Prometheus docs - each label value on a Counter is an individual object that will live and wait around to be incremented (and take up resources). I'm going to try initializing them all on startup, I think this is the right way to use Counters ([reference](https://prometheus.github.io/client_python/instrumenting/labels/)). Additionally, the docs say a Counter shouldn't have more than 10 labels ([reference](https://prometheus.io/docs/practices/instrumentation/#use-labels)), and this one is >50, so we will need to monitor prometheus performance around this deploy.

### DIFF
--- a/shared/torngit/base.py
+++ b/shared/torngit/base.py
@@ -50,33 +50,6 @@ class TorngitBaseAdapter(object):
         "xtend",
     )
 
-    def get_client(self, timeouts: List[int] = []) -> httpx.AsyncClient:
-        if timeouts:
-            timeout = httpx.Timeout(timeouts[1], connect=timeouts[0])
-        else:
-            timeout = httpx.Timeout(self._timeouts[1], connect=self._timeouts[0])
-        return httpx.AsyncClient(
-            verify=self.verify_ssl
-            if not isinstance(self.verify_ssl, bool)
-            else self.verify_ssl,
-            timeout=timeout,
-        )
-
-    def get_token_by_type(self, token_type: TokenType):
-        if self._token_type_mapping.get(token_type) is not None:
-            return self._token_type_mapping.get(token_type)
-        return self.token
-
-    def get_token_by_type_if_none(self, token: Optional[str], token_type: TokenType):
-        if token is not None:
-            return token
-        return self.get_token_by_type(token_type)
-
-    def _oauth_consumer_token(self):
-        if not self._oauth:
-            raise Exception("Oauth consumer token not present")
-        return self._oauth
-
     def __init__(
         self,
         oauth_consumer_token: OauthConsumerToken = None,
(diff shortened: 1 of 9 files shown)


shared/torngit/base.py
```diff
@@ -107,6 +80,33 @@ def __repr__(self):
...
+    def get_token_by_type_if_none(self, token: Optional[str], token_type: TokenType):
+        if token is not None:
+            return token
+        return self.get_token_by_type(token_type)
+
+    def _oauth_consumer_token(self):
+        if not self._oauth:
+            raise Exception("Oauth consumer token not present")
+        return self._oauth
+
```
nora-codecov, line 109: No changes in function here, just organization

shared/torngit/github.py
```diff
@@ -154,30 +425,30 @@ def get_api_url(cls):
...
         return self._token
 
-    @property
-    def api_host_header(self):
-        return self.get_api_host_header()
-
-    @property
-    def host_header(self):
-        return self.get_host_header()
-
```
nora-codecov, line 180: No functional changes to these `@property` methods, just organization

shared/torngit/github.py
```diff
@@ -37,6 +38,292 @@
...
+        "url_template": Template("/app/hook/deliveries/${delivery_id}/attempts"),
+    },
+    "refresh_token": {
+        "counter": GITHUB_API_CALL_COUNTER.labels(endpoint="refresh_token"),
+        "url_template": Template("/login/oauth/access_token"),
+    },
+    "make_http_call_retry": {
+        "counter": GITHUB_API_CALL_COUNTER.labels(endpoint="make_http_call_retry"),
+        "url_template": "",  # no url template, just counter
+    },
```
matt-codecov, line 62: nice

matt-codecov (approved):
awesome work! really excited to have these metrics!

PR #186, opened 2024-04-15
feat: improve fallback system

giovanni-guidini:
These changes improve the github token fallback system by:
1. Updating the instance's `self._token` to be a not-rate-limited one,
  so that if the instance make subsequent requests they won't be immediately
  rate limited and have to go through the fallback process again
2. Propagates to other parts of the system that a given installation is rate limited.
  This is done via redis (apparently shared can access redis now)

Sadly the function that fallsbakc if other tokens are available is full of
side effects, as documented.

Also did a small refactor in the code aggregating the primary and secondary rate limit errors.

### Legal Boilerplate

### DIFF
--- a/shared/torngit/github.py
+++ b/shared/torngit/github.py
@@ -1,5 +1,6 @@
 import asyncio
 import base64
+import datetime
 import hashlib
 import logging
 import os
@@ -12,10 +13,10 @@ import httpx
 from httpx import Response
 
 from shared.config import get_config
-from shared.github import get_github_jwt_token
+from shared.github import get_github_jwt_token, mark_installation_as_rate_limited
 from shared.metrics import Counter, metrics
 from shared.torngit.base import TokenType, TorngitBaseAdapter
-from shared.torngit.cache import torngit_cache
+from shared.torngit.cache import get_redis_connection, torngit_cache
 from shared.torngit.enums import Endpoints
 from shared.torngit.exceptions import (
     TorngitClientError,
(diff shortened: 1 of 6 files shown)


shared/torngit/github.py
```diff
@@ -518,16 +527,71 @@ def _parse_response(self, res: Response):
...
+    def _get_next_fallback_token(
+        self,
+        *,
+        reset_timestamp: Optional[str] = None,
+        retry_in_seconds: Optional[int] = None,
+    ) -> Optional[str]:
         """If additional fallback tokens were passed to this instance of GitHub
         select the next token in line to retry the previous request.
+
+        !side effect: Marks the current token as rate limited in redis
```
joseph-sentry, line 571: I like these side effect comments!

shared/torngit/github.py
```diff
@@ -518,16 +527,71 @@ def _parse_response(self, res: Response):
...
+                log.warning(
+                    "Can't mark installation as rate limited because TTL is missing",
+                    extra=dict(installation_id=installation_id),
+                )
+                return
+            ttl_seconds = retry_in_seconds
+            if ttl_seconds is None:
+                # https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#handle-rate-limit-errors-appropriately
+                ttl_seconds = int(reset_timestamp) - int(
+                    datetime.datetime.now(datetime.timezone.utc).timestamp()
```
joseph-sentry, line 549: [nit] this line is a bit hard to read, I think it might be easier to read if we imported what we were using from datetime only
```
from datetime import datetime, timezone
```

joseph-sentry (approved):
looks good, just left one nit comment

nora-codecov (approved):

PR #188, opened 2024-04-15
fix: Sessions table not populating

JerrySentry:
Add a link between `django_session` and `codecov_auth.sessions` table. This allows Django's auth module to create an object in the Codecov sessions table to track login metadata. The sessions table on the UI shows the IP, user agent, login time. This fixes the issue where `codecov_auth.sessions` was not populating ever since migrating to Django's auth mechanism. The second part of the fix will be on the api side.

### Legal Boilerplate


shared/django_apps/codecov_auth/models.py
```diff
@@ -7,6 +7,7 @@
 from hashlib import md5
 
 from django.contrib.postgres.fields import ArrayField, CITextField
+from django.contrib.sessions.models import Session
```
rohitvinnakota-codecov, line 10: Can we import as `DjangoSession` just in case of any circular references below?

rohitvinnakota-codecov (approved):

PR #189, opened 2024-04-15
fix: do bulk inserts in bundle analysis parser

joseph-sentry:
We were originally doing inserts every time we created an Asset, Chunk, or Module object. Now we are doing the inserts into the DB after having created them all.

Fixes: https://github.com/codecov/internal-issues/issues/399

### DIFF
--- a/shared/bundle_analysis/parser.py
+++ b/shared/bundle_analysis/parser.py
@@ -30,16 +30,10 @@ class Parser:
         """
         Resets temporary parser state in order to parse a new file path.
         """
-        # asset name -> asset id
-        self.asset_index = {}
-
-        # chunk external id -> chunk internal id
-        self.chunk_index = {}
-
-        # chunk id -> asset name list
+        # chunk unique id -> asset name list
         self.chunk_asset_names_index = {}
 
-        # module id -> chunk external id list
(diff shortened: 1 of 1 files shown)


shared/bundle_analysis/parser.py
```diff
@@ -215,30 +243,49 @@ def _parse_modules_event(self, prefix: str, event: str, value: str):
...
-            assets = self.db_session.query(Asset).filter(
-                Asset.session == self.session,
-                Asset.name.in_(asset_names),
-            )
+        assets: list[Asset] = self.db_session.query(Asset).filter(
+            Asset.session_id == self.session.id,
+        ).all()
+
+        asset_name_to_id = {
+            asset.name: asset.id for asset in assets
```
giovanni-guidini, line 251: Is it possible to have multiple assets with the same name? Would that overwrite some IDs?

joseph-sentry: it is not possible, but if it was, you're correct that this solution would be broken

giovanni-guidini: It would be helpful to add a comment stating that

giovanni-guidini (approved):

PR #190, opened 2024-04-16
feat: prevent the app from destroying itself

giovanni-guidini:
in a few occasions we had in staging (and 1 cloud customer) scenarios that
lead us to believe that the default app_id at times might not be present for the api pods
config.

The bad side_effect is that installations change name to 'unconfigured_app' and that's
only manually reversible. To prevent the app from destroying itself if the default
app id is not present we assume apps are configured in this scenario.

### Legal Boilerplate

### DIFF
--- a/tests/unit/django_apps/codecov_auth/test_codecov_auth_models.py
+++ b/tests/unit/django_apps/codecov_auth/test_codecov_auth_models.py
@@ -595,7 +595,7 @@ class TestGithubAppInstallationModel(TransactionTestCase):
         )
 
 
-class TestGitguvAppInstallationNoDefaultAppIdConfig(TransactionTestCase):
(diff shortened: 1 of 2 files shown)


tests/unit/django_apps/codecov_auth/test_codecov_auth_models.py
```diff
@@ -592,3 +593,21 @@ def test_is_configured(self):
         assert (
             installation_default_name_not_default_id_configured.is_configured() == True
         )
+
+
+class TestGitguvAppInstallationNoDefaultAppIdConfig(TransactionTestCase):
```
thomasrockhu-codecov, line 598: typo?

```suggestion
class TestGitHubAppInstallationNoDefaultAppIdConfig(TransactionTestCase):
```

joseph-sentry (approved):

PR #197, opened 2024-04-24
feat: add commitid and branch to testinstance

joseph-sentry:
- create migration to add commitid and branch nullable fields to TestInstance to avoid having to do joins to get that information

- create migration to set value of commitid and branch for all TestInstance objects to match commitid and branch of Commit associated with TestInstance


shared/django_apps/reports/migrations/0018_testinstance_branch_testinstance_commitid.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.11 on 2024-04-24 18:22
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
```
giovanni-guidini, line 7: [kinda nit] 
Typically we run `python manage.py sqlmigrate <APP_NAME> <MIGRATION_NUMBER>`
and paste the SQL output as comments inside the migration, just to have a better idea of the SQL that's gonna run for the migration

giovanni-guidini (commented):
LGTM

PR #199, opened 2024-04-25
Add count_and_get_url_template to gitlab handler

nora-codecov:
Part of https://github.com/codecov/engineering-team/issues/1167, this is for GitLAB calls, as I continue with the other git providers.

We want to know who's making all these calls! So I added some Counters.

### DIFF
--- a/shared/torngit/gitlab.py
+++ b/shared/torngit/gitlab.py
@@ -2,13 +2,14 @@ import json
 import logging
 import os
 from base64 import b64decode
+from string import Template
 from typing import List
 from urllib.parse import quote, urlencode
 
 import httpx
 
 from shared.config import get_config
-from shared.metrics import metrics
+from shared.metrics import Counter, metrics
 from shared.torngit.base import TokenType, TorngitBaseAdapter
 from shared.torngit.enums import Endpoints
 from shared.torngit.exceptions import (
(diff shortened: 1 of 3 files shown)


shared/torngit/gitlab.py
```diff
@@ -29,24 +30,242 @@
...
+
+
+GITLAB_API_ENDPOINTS = {
+    "fetch_and_handle_errors_retry": {
+        "counter": GITLAB_API_CALL_COUNTER.labels(
+            endpoint="fetch_and_handle_errors_retry"
+        ),
+        "url_template": "",  # no url template, just counter
+    },
+    "get_best_effort_branches": {
```
adrian-codecov, line 47: [optional] Thoughts on making these keys part of an ENUM? For instance, I've seen "get_best_effort_branches" mentioned 4 times in here, so would centralize the naming if we ever change this. Not a huge pain to find/replace those 4 instances though, whatever you think is better

adrian-codecov (approved):
Left a little comment but otherwise lgtm!

I'm curious if we could wrap the count_and_get_url_template logic in a decorator 👀. I suppose you still need to return the url, but would be neat to have all fns that count wrapped in one single line 👀 food for thought

matt-codecov (approved):

PR #202, opened 2024-04-25
update the shared django settings

adrian-codecov:
Updating the shared db django settings to be used in worker/api

### Legal Boilerplate

### DIFF
--- a/shared/django_apps/db_settings.py
+++ b/shared/django_apps/db_settings.py
@@ -112,7 +112,7 @@ CONN_MAX_AGE = int(get_config("services", "database", "conn_max_age", default=0)
 
 DATABASES = {
     "default": {
-        "ENGINE": "django.db.backends.postgresql",
+        "ENGINE": "psqlextra.backend",
         "NAME": DATABASE_NAME,
         "USER": DATABASE_USER,
         "PASSWORD": DATABASE_PASSWORD,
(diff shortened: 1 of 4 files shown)


shared/torngit/github.py
```diff
@@ -1501,9 +1501,9 @@ async def post_webhook(self, name, url, events, secret, token=None):
         token = self.get_token_by_type_if_none(token, TokenType.admin)
         # https://developer.github.com/v3/repos/hooks/#create-a-hook
         async with self.get_client() as client:
-            api_url = self.count_and_get_url_template(url_name="post_webhook").substitute(
-                slug=self.slug
-            )
+            api_url = self.count_and_get_url_template(
```
adrian-codecov, line 1504: Linting

shared/django_apps/db_settings.py
```diff
@@ -112,7 +112,7 @@
 
 DATABASES = {
     "default": {
-        "ENGINE": "django.db.backends.postgresql",
+        "ENGINE": "psqlextra.backend",
```
adrian-codecov, line 115: Updating these and not just dummy_settings.py so they are picked up by api + worker

shared/django_apps/legacy_migrations/management/commands/migrate.py
```diff
@@ -99,6 +99,7 @@ def handle(self, *args, **options):
                 self._fake_initial_migrations(cursor, args, options)
 
             super().handle(*args, **options)
+            django_transaction.commit(database)
```
adrian-codecov, line 102: Explicitly commit as children apps that inherit this might not (like worker)

shared/torngit/github.py
nora-codecov: That's weird! sorry about that!

nora-codecov (approved):

shared/torngit/github.py
adrian-codecov: haha don't worry!!

PR #205, opened 2024-04-29
feat: add more model fields + refactor feature interface

daniel-codecov:
Adds more fields to `FeatureFlag` and `FeatureFlagVariant`. Primarily prep work so that we can expose flags via an endpoint on codecov-api for gazebo to use.

`platform` - either FRONTEND or BACKEND. We make this distinction so we can tell on django admin whether a feature flag is for frontend or backend. Frontend flags don't have 
`is_active` - if the experiment is running currently, or it is off/cleaned up. Just for us to keep things tidy
`rollout_identifier` - define which field we're rolling out over. Before we have just owner_id and repo_id, but now we've added email and org_id

The current flags in prod don't have these fields, but the default values for those fields shouldn't introduce anything breaking. When we deploy this, we need to manually update the current flags to the appropriate new values.

### Legal Boilerplate


shared/django_apps/rollouts/models.py
```diff
@@ -1,17 +1,31 @@
...
+
+# Defines the possible identifiers you can perform a rollout over. Users
+# who share the same identifier will always be assigned the same variant.
+# EG: two users from the same org will receive the same variant when rolling
+# out over ORG_ID
+class RolloutIdentifier(models.TextChoices):
+    OWNER_ID = "U", "Owner ID"
+    REPO_ID = "R", "Repo ID"
+    ORG_ID = "O", "Org ID"
+    EMAIL = "E", "Email"
```
matt-codecov, line 21: using "U" for OWNER_ID is a little awkward because we also have a `User` model which we could hypothetically also want to run an experiment over. a `User` is a user of codecov, and an `Owner` is an entity that can own repos on a git provider (so a user or org on GitHub, GitLab, or Bitbucket)

i think there will be few enough rows that we can use full words for these choices and it won't cost much space. also is there a benefit to having orgs and owners be separate?

shared/django_apps/rollouts/models.py
```diff
@@ -28,6 +42,24 @@ class FeatureFlag(models.Model):
...
+    )
+
+    # The field we're rolling out over. Users with the same identifier
+    # will always receive the same variant. EG: if you rollout over org_id,
+    # then users in the same org see the same variant
+    rollout_identifier = models.CharField(
+        max_length=1,
+        choices=RolloutIdentifier.choices,
+        default=RolloutIdentifier.OWNER_ID,
+    )
```
matt-codecov, line 62: i think `rollout_universe` feels a little clearer to me than `rollout_identifier` but not a strong preference

shared/django_apps/rollouts/models.py
daniel-codecov: Yeah I can just make it the whole string then

Also, benefit of owners and orgs separate is just for clarity. I don't feel strongly about either decision to merge or keep them separate. Is it common knowledge that orgs are also owners?

shared/django_apps/rollouts/models.py
```diff
@@ -28,6 +42,24 @@ class FeatureFlag(models.Model):
...
+    # is no longer running anymore
+    is_active = models.BooleanField(
+        default=True,
+        help_text="This should be on if the experiment is currently running. Otherwise turn it off if the experiment has finished and is cleaned up",
+    )
+
+    # The field we're rolling out over. Users with the same identifier
+    # will always receive the same variant. EG: if you rollout over org_id,
+    # then users in the same org see the same variant
+    rollout_identifier = models.CharField(
```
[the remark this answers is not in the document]
daniel-codecov: thanks

matt-codecov (approved):

PR #207, opened 2024-04-29
fix: typing issue because api is on 3.9 still

daniel-codecov:
### Legal Boilerplate


shared/bundle_analysis/parser.py
```diff
@@ -74,21 +74,15 @@ def parse(self, path: str) -> int:
                     self._parse_event(event)
 
                 if self.asset_list:
-                    insert_asset = Asset.__table__.insert().values(
```
daniel-codecov, line 77: lint

giovanni-guidini (approved):

PR #216, opened 2024-05-07
Update 0019_auto_20240424_1824.py

trent-codecov:
### Legal Boilerplate


adrian-codecov (approved):

shared/django_apps/reports/migrations/0019_auto_20240424_1824.py
```diff
@@ -25,6 +25,8 @@ class Migration(migrations.Migration):
     ]
 
     def populate_test_instances(apps, schema_editor):
+        print("Not running due to performance")
```
thomasrockhu-codecov, line 28: lolol

PR #219, opened 2024-05-10
dev: Add Ruff to Shared, enable F541 and "I" rules

ajay-sentry:
Copy of https://github.com/codecov/codecov-api/pull/555, but for shared.

### Legal Boilerplate


ruff.toml
```diff
@@ -0,0 +1,61 @@
...
+# Same as Black.
+line-length = 88
+indent-width = 4
+
+# Assume Python 3.10
+target-version = "py310"
+
+[lint]
+# Currently only enabled F541: https://docs.astral.sh/ruff/rules/
+select = ["F541", "I"]
```
ajay-sentry, line 40: There is 1 difference here between the API PR and this one, which is that we are also enabling the isort rules
https://docs.astral.sh/ruff/rules/unsorted-imports/

rohitvinnakota-codecov (approved):

PR #222, opened 2024-05-16
fix: default branch to "master" in torngit github

matt-codecov:
fixes a sentry issue https://l.codecov.dev/I8iSfU

we upsert a repo with `branch=None` due to this default value. that violates the nonnull constraint on that field

other places in the same file default to "master" as well: [one](https://github.com/codecov/shared/blob/5ada529179d25063f02e64be0a9cba3129d8d5d6/shared/torngit/github.py#L1175) and [two](https://github.com/codecov/shared/blob/5ada529179d25063f02e64be0a9cba3129d8d5d6/shared/torngit/github.py#L1159)

### DIFF
--- a/shared/torngit/github.py
+++ b/shared/torngit/github.py
@@ -1285,7 +1285,7 @@ class Github(TorngitBaseAdapter):
                         ),
                         "private": raw_repo_data["isPrivate"],
                         "branch": (
-                            default_branch.get("name") if default_branch else None
+                            default_branch.get("name") if default_branch else "master"
                         ),
                         "owner": {
                             "node_id": raw_repo_data["owner"]["id"],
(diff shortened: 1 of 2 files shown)


nora-codecov (approved):

shared/torngit/github.py
```diff
@@ -1285,7 +1285,7 @@ async def get_repos_from_nodeids_generator(
                         ),
                         "private": raw_repo_data["isPrivate"],
                         "branch": (
-                            default_branch.get("name") if default_branch else None
+                            default_branch.get("name") if default_branch else "master"
```
adrian-codecov, line 1288: Wouldn't we want to do main?

matt-codecov: other places in this code do "master" and i thought i should be consistent

PR #224, opened 2024-05-21
Increase the decimal precision to 5 decimal places for reports

michelletran-codecov:
This is a risky migration that increase the number of decimal places for the `uploadleveltotals` and the `reportleveltotals` tables. This is to fix the rounding issues when displaying the results on the UI.

Marking this as a risky migration because the `uploadleveltotals` table is 65Gb and the `reportleveltotals` table is 15Gb. 

ref: https://github.com/codecov/engineering-team/issues/1614

### Legal Boilerplate


adrian-codecov (approved):
Lgtm! If you're unfamiliar w/ the process to migrate risky migrations hmu on slack and I can guide you through!

PR #227, opened 2024-05-24
Bundle Analysis: update asset model

JerrySentry:
Adds 2 columns to the asset table:
1. uuid -> this is for associating the current asset with an asset in the previous commit. when two assets with the same uuid is the same across two commits then they are considered the same.
2. asset_type -> store the type determined via extensions, different types will need to be processed differently moving forward.

closes: https://github.com/codecov/engineering-team/issues/1767

### Legal Boilerplate


shared/bundle_analysis/utils.py
```diff
@@ -0,0 +1,21 @@
...
+
+def get_extension(filename: str) -> str:
+    """
+    Gets the file extension of the file without the dot
+    """
+    # At times file can be something like './index.js + 12 modules', only keep the real filepath
+    filename = filename.split(" ")[0]
+    # Retrieve the file extension with the dot
+    _, file_extension = os.path.splitext(filename)
+    # Return empty string if file has no extension
```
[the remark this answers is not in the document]
JerrySentry: Good catch, updated.

shared/bundle_analysis/utils.py
```diff
@@ -0,0 +1,21 @@
...
+    if not file_extension or file_extension[0] != ".":
+        return file_extension
+    # Remove the dot in the extension
+    file_extension = file_extension[1:]
+    # At times file can be something like './index.js?module', remove the ?
+    if "?" in file_extension:
+        file_extension = file_extension[: file_extension.rfind("?")]
+
```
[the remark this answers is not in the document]
JerrySentry: Ah true, using this.

shared/bundle_analysis/parser.py
```diff
@@ -183,6 +200,8 @@ def _parse_assets_event(self, prefix: str, event: str, value: str):
                     name=self.asset.name,
                     normalized_name=self.asset.normalized_name,
                     size=self.asset.size,
+                    uuid=str(uuid.uuid4()),
```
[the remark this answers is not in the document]
JerrySentry: You are right, this is just the part 1 of the change to make linking assets across different commits work. By default we will create the `uuid` for an asset for this PR, for the next PR as the next step of the parsing process it will look at the previous commit's asset, if the algorithm thinks its the same then it will update the current asset's `uuid` to be the same as the previous, otherwise it will remain the same.

I choose to initialize an `uuid` on creation and then update it in the following parsing step is because the parsing code is already kind of convoluted, so setting the final `uuid` right now it would be a nightmare.

shared/bundle_analysis/utils.py
```diff
@@ -0,0 +1,21 @@
...
+    filename = filename.split(" ")[0]
+    # Retrieve the file extension with the dot
+    _, file_extension = os.path.splitext(filename)
+    # Return empty string if file has no extension
+    if not file_extension or file_extension[0] != ".":
+        return ""
+    # Remove the dot in the extension
+    file_extension = file_extension[1:]
+    # At times file can be something like './index.js?module', remove the ?
+    if "?" in file_extension:
```
ajay-sentry, line 18: isn't this doing 2 passes for essentially the same thing? At least in JS, if we just did file_extension.split("?")[0] and there was no "?" it would return the same thing so I figure python would be the same.
Screenshot 2024-05-28 at 1 46 43 PM

shared/bundle_analysis/parser.py
```diff
@@ -123,6 +126,20 @@ def parse(self, path: str) -> int:
             e.bundle_analysis_plugin_name = self.info.get("plugin_name", "unknown")
             raise e
 
+    def _asset_type(self, name: str) -> AssetType:
+        extension = get_extension(name)
+
+        if extension in ["js"]:
```
ajay-sentry, line 132: just curious, where did we compile this list of extensions?

shared/bundle_analysis/utils.py
JerrySentry: Aha true! updated.

shared/bundle_analysis/parser.py
JerrySentry: this matches how Sentry buckets its asset types.
Or defined here https://www.notion.so/sentry/Type-Filtering-and-Grouping-82d5db9fd7f84737a4792df2ae30c22f

ajay-sentry: perf! Thanks, exactly what I was looking for

ajay-sentry (approved):
lgtm!

PR #228, opened 2024-05-27
Update repo model with tests enabled

RulaKhaled:
Needed to conditionally render onboarding for failed tests.

### Legal Boilerplate


shared/django_apps/core/models.py
```diff
@@ -139,6 +139,7 @@ class Languages(models.TextChoices):
     deleted = models.BooleanField(default=False)
     bundle_analysis_enabled = models.BooleanField(default=False, null=True)
     coverage_enabled = models.BooleanField(default=False, null=True)
+    tests_enabled = models.BooleanField(default=False, null=True)
```
nicholas-codecov, line 142: Can we slightly change this to be `test_analytics_enabled` just to be super clear as to what we're talking about?

RulaKhaled: ✅

shared/django_apps/core/migrations/0050_repository_tests_enabled.py
```diff
@@ -0,0 +1,27 @@
...
+    ALTER TABLE "repos" ALTER COLUMN "test_analytics_enabled" DROP DEFAULT;
+    COMMIT;
+    """
+
+    dependencies = [
+        ("core", "0049_increment_version"),
+    ]
+
+    operations = [
+        migrations.AddField(
```
JerrySentry, line 22: let's use the RiskyAddField for this since its updating the repos table

RulaKhaled: interesting, anything we add to repo is a risky add?

JerrySentry: Generally yes (similar with other widely used tables like owners and commits) because the migration will timeout because it will be blocked for a long time by other operations touching the table, but if your new column is `(default=False, null=True)` then in theory it should be ok. But I'm not sure that's always the case and past migrations have gotten me paranoid. So IMO just be safe and do risky.

JerrySentry (approved):

shared/django_apps/core/migrations/0050_repository_tests_enabled.py
adrian-codecov: Just in case, the "new" way we get these into prod is by merging this code and then updating the shared version in api/worker. When you do so, schedule sometime with the infra team to run the migration after hours, cause if you merge the code itself, the migrations won't be applied, so they need to run the SQL for you then.

adrian-codecov (dismissed):

shared/django_apps/core/migrations/0050_repository_tests_enabled.py
RulaKhaled: Thank you both!

adrian-codecov (approved):

PR #231, opened 2024-05-29
Bundle Analysis: associate past assets to current parsed bundle

JerrySentry:
Asset name changes from one bundler build to another across commits, therefore we don't know if the current asset is a new one or a continuation of the previous one. This PR adds a function to offer a heuristic to associate the current asset to an asset from the previous commit.

The rules are that if the hashed asset name exists in the previous bundle report then this is considered the same asset. Similarly if all the modules of the asset are the same as any in the previous bundle report it is also considered the same asset. We track assets through a generated UUID, when 2 assets are considered associated then they will have the same UUID.

This mechanism is important for providing analytics asset size trends throughout the course of its existence. This component will be implemented in the coming iterations. 

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/report.py
+++ b/shared/bundle_analysis/report.py
@@ -171,10 +171,10 @@ class BundleAnalysisReport:
         self.db_session.commit()
         return session_id
 
-    def associate_previous_assets(self, prev_bundle_analysis_report: Any) -> None:
+    def associate_previous_assets(
+        self, prev_bundle_analysis_report: Any
+    ) -> "BundleAnalysisReport":
         """
-        Note: prev_bundle_analysis_report is of type BundleAnalysisReport,
-              typing.Self is not available in 3.10
         Only associate past asset if it is Javascript or Typescript types
         and belonging to the same bundle name
         Associated if one of the following is true
@@ -232,6 +232,7 @@ class BundleAnalysisReport:
 
                     # Update the Assets table for the bundle
                     # TODO: Use SQLalchemy ORM to update instead of raw SQL
+                    # https://github.com/codecov/engineering-team/issues/1846
                     for pair in associated_assets_found:
                         prev_uuid, curr_uuid = pair
                         stmt = f"UPDATE assets SET uuid='{prev_uuid}' WHERE uuid='{curr_uuid}'"
--- a/shared/bundle_analysis/report.py
+++ b/shared/bundle_analysis/report.py
@@ -2,8 +2,9 @@ import json
 import os
 import sqlite3
 import tempfile
-from typing import Any, Dict, Iterator, Optional
+from typing import Any, Dict, Iterator, Optional, Self
 
+from sqlalchemy import text
 from sqlalchemy.exc import OperationalError
 from sqlalchemy.sql import func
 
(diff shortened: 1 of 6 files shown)


shared/bundle_analysis/report.py
```diff
@@ -170,6 +171,73 @@ def ingest(self, path: str) -> int:
...
+                            if curr_modules in prev_module_asset_mapping:
+                                associated_assets_found.append(
+                                    [
+                                        prev_module_asset_mapping[curr_modules],
+                                        curr_asset.uuid,
+                                    ]
+                                )
+
+                    # Update the Assets table for the bundle
+                    # TODO: Use SQLalchemy ORM to update instead of raw SQL
```
giovanni-guidini, line 234: Is there a ticket for this TODO already? Big chances of it being forgotten

tests/unit/bundle_analysis/test_asset_association.py
```diff
@@ -0,0 +1,92 @@
...
+def _get_asset_mapping(
+    bundle_analysis_report: BundleAnalysisReport, bundle_name: Asset
+) -> Dict[str, str]:
+    bundle_report = bundle_analysis_report.bundle_report(bundle_name)
+    asset_report = list(bundle_report.asset_reports())
+    return {asset.hashed_name: asset for asset in asset_report}
+
+
+def test_asset_association():
+    try:
```
giovanni-guidini, line 38: Why have a `try:` without an `except`?

"to have the `finally;` block" 
but what's the difference of having it at the end of the try block?

giovanni-guidini (approved):
I personally think that `associate_previous_assets` is quite complex as it is. I'd encourage you to refactor it by breaking up parts of the big loop into helper functions.
Easier to read and to test. But the comments and docstrings are quite helpful, thanks for that.

Otherwise LGTM... left some other comments.

shared/bundle_analysis/report.py
JerrySentry: Added link to it.

shared/bundle_analysis/report.py
```diff
@@ -170,6 +171,73 @@ def ingest(self, path: str) -> int:
         self.db_session.commit()
         return session_id
 
+    def associate_previous_assets(self, prev_bundle_analysis_report: Any) -> None:
+        """
+        Note: prev_bundle_analysis_report is of type BundleAnalysisReport,
+              typing.Self is not available in 3.10
```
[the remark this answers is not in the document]
JerrySentry: Oh interesting I didn't know about the double quote trick!

giovanni-guidini (approved):

JerrySentry:
> I personally think that `associate_previous_assets` is quite complex as it is. 

Separated out the 2 rules as 2 separate helper functions

shared/bundle_analysis/report.py
```diff
@@ -170,6 +171,104 @@ def ingest(self, path: str) -> int:
         self.db_session.commit()
         return session_id
 
+    def _associate_bundle_report_assets_by_name(
+        self, curr_bundle_report, prev_bundle_report
+    ) -> Set:
```
giovanni-guidini, line 176: [very nit] You can be more explicit by saying `Set[Tuple]` or even more `Set[Tuple[str, str]]`. Starts to get a bit wild fast but possible. I personally think `Set[Tuple]` is an improvement over `Set`.

shared/bundle_analysis/report.py
```diff
@@ -170,6 +171,104 @@ def ingest(self, path: str) -> int:
         self.db_session.commit()
         return session_id
 
+    def _associate_bundle_report_assets_by_name(
+        self, curr_bundle_report, prev_bundle_report
```
giovanni-guidini, line 175: The args don't have a known type?

shared/bundle_analysis/report.py
```diff
@@ -170,6 +171,104 @@ def ingest(self, path: str) -> int:
...
+        Rule 1. Previous and current asset have the same hashed name
+        Rule 2. Previous and current asset shared the same set of module names
+        """
+        for curr_bundle_report in self.bundle_reports():
+            for prev_bundle_report in prev_bundle_analysis_report.bundle_reports():
+                if curr_bundle_report.name == prev_bundle_report.name:
+                    associated_assets_found = set()
+
+                    # Rule 1 check
+                    associated_assets_found |= (
```
giovanni-guidini, line 250: Even the code is going "|=" (i.e. 😐)
(Jokes aside I do think this is an improvement, thanks)

giovanni-guidini (approved):

giovanni-guidini (approved):

PR #233, opened 2024-05-30
torngit github: retry on 502/503/504

matt-codecov:
https://github.com/codecov/engineering-team/issues/1757

we sometimes observe 502s and 504s in `list_repos()`. https://github.com/codecov/shared/pull/232 attempts to mitigate that by testing a smaller page size, but also sometimes retrying 502/503/504 will just work


adrian-codecov (approved):
Curious to see how this looks, I'd pay close attention to undesired side effects

PR #235, opened 2024-05-31
dev: Remove all the excluded ruff rules except E501

ajay-sentry:
Part of the lint enhancement epic: https://github.com/codecov/engineering-team/issues/1716

This PR aims to fix all the "hanging" F and E rules that we excluded previously. The rules that we are now enforcing are as follows:

- F401: {name} imported but unused; consider using importlib.util.find_spec to test for availability
- F403: from {name} import * used; unable to detect undefined names
- F405: {name} may be undefined, or defined from star imports
- F841: Local variable {name} is assigned to but never used
- E711: Comparison to None should be cond is None
- E712: Avoid equality comparisons to True; use if {cond}: for truth checks
- E741: Ambiguous variable name: {name}

Rules List: https://docs.astral.sh/ruff/rules/#legend

This leaves one remaining rule, E501, which is about line length unchanged. I think for now it doesn't make sense to add this as an enforced rule since it's more stylistic, but we can always consider turning it on later.

Closes https://github.com/codecov/engineering-team/issues/1852

### Legal Boilerplate


ruff.toml
```diff
@@ -26,6 +26,7 @@ exclude = [
     "node_modules",
     "site-packages",
     "venv",
+    "**test**"
```
ajay-sentry, line 29: One thing too, since there were many, many errors in the test files was I just excluded all of them for the sake of time. If someone wants to go back and fix those as well by all means :)

shared/bundle_analysis/__init__.py
```diff
@@ -16,3 +16,21 @@
     ModuleReport,
 )
 from shared.bundle_analysis.storage import BundleAnalysisReportLoader, StoragePaths
+
+__all__ = [
+    "models",
```
ajay-sentry, line 21: This is how we can get around the "unused imports" for these init files

shared/torngit/base.py
```diff
@@ -219,8 +219,8 @@ def diff_to_json(self, diff):
                 elif sol4 == "@@ -":
                     # ex: "@@ -31,8 +31,8 @@ blah blah blah"
                     # ex: "@@ -0,0 +1 @@"
-                    l = get_start_of_line(source).groups()
```
ajay-sentry, line 222: "l" is an ambiguous character, and rightfully so. changed to "ln" to signal "line" which is what "l" was originally representing

shared/torngit/bitbucket.py
```diff
@@ -802,7 +802,7 @@ async def find_pull_request(
         self, commit=None, branch=None, state="open", token=None
     ):
         state = {"open": "OPEN", "merged": "MERGED", "close": "DECLINED"}.get(state, "")
-        pulls, page = [], 0
```
ajay-sentry, line 805: pulls wasn't being populated here, just an empty array

JerrySentry (approved):
lgtm

PR #240, opened 2024-06-05
[feat] Add user onboarding lifecycle table

rohitvinnakota-codecov:
https://www.notion.so/sentry/Eng-Design-Tracking-Codecov-user-signup-to-first-commit-conversion-rate-6097411516f84563820276aafe3436ea

Adding a table where we can track events coming in from different services to better understand user onboarding flow. 

This PR closes https://github.com/codecov/engineering-team/issues/1828

### Legal Boilerplate


shared/django_apps/core/models.py
```diff
@@ -508,3 +508,10 @@ class CommitError(ExportModelOperationsMixin("core.commit_error"), BaseCodecovMo
 
     class Meta:
         app_label = CORE_APP_LABEL
+
+
+class UserOnboardingLifeCycleMetrics(models.Model):
```
adrian-codecov, line 513: Thoughts on making this a standalone django app? Or ist here a strong reason this should be in `core`?

shared/django_apps/core/models.py
```diff
@@ -508,3 +508,10 @@ class CommitError(ExportModelOperationsMixin("core.commit_error"), BaseCodecovMo
...
+    id = models.BigAutoField(primary_key=True)
+    orgid = models.IntegerField()
```
adrian-codecov, line 515: Nit: orgid looks a bit weird to me, thoughts on `org_id`?

shared/django_apps/core/migrations/0053_useronboardinglifecyclemetrics.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 4.2.13 on 2024-06-07 14:57
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
```
adrian-codecov, line 7: Could you pls add the sql comment here? See other migrations for reference, you'd need to run `python manage.py sqlmigrate <app> <number>`; `python manage.py sqlmigrate core 0053`

shared/django_apps/core/models.py
rohitvinnakota-codecov: I can do that

adrian-codecov (approved):

PR #243, opened 2024-06-07
fix: Check for visibility of emails - GitHub 

RulaKhaled:
We fetch user's email in `/user` which according to GitHub is "The first primary publicly visible email" see [here](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-a-user), if that does not exist we should be fetching the first visible email, instead of ignoring the visibility of the primary email. 

resources: https://docs.github.com/en/rest/users/emails?apiVersion=2022-11-28#list-email-addresses-for-the-authenticated-user

Fixes: https://github.com/codecov/feedback/issues/391 

### Legal Boilerplate


tests/integration/test_github.py
```diff
@@ -205,7 +205,7 @@ async def test_get_authenticated_user_no_refresh_token(self, codecov_vcr):
             "company": "@codecov ",
             "blog": "",
             "location": None,
-            "email": "thiago@codecov.io",
+            "email": None,
```
adrian-codecov, line 208: Mmm is there a test where we actually get the email and verify it's public?

ajay-sentry: I was also thinkin, how can we be certain we're still fetching the primary public email? 
Maybe we could check if the primary is public and return, otherwise find any public?

RulaKhaled: > how can we be certain we're still fetching the primary public email

@ajay-sentry that's what we're doing! email in `/user` defaults to the fist public primary email (it's in the docs), if it doesn't exist we fetch the first public email in `/emails`. That's assuring us we're respecting the "visibility" of emails. (resources in the desc)

@adrian-codecov tbh i'm not sure, this fetches `https://github.com/login/oauth/authorize?response_type=code&scope=user%3Aemail%2Cread%3Aorg%2Crepo%3Astatus%2Cwrite%3Arepo_hook&client_id=999247146557c3ba045c` and tests it's values, i'm not sure how i can alter values in such case, any thoughts?

adrian-codecov: Are you familiar w/ cassettes? 

The idea is that you make a real API call once, that creates a cassette, and then your test calls that cassette instead of calling the actual API. So you'd want to make a brand new test to an org you have a token of /access to. You'll need to use one the `*_handlers` , and replace the token with a real token/owner/repo combo for the test, and have that create a cassette. Does that make sense? Lmk and I can give you a hand otherwise

RulaKhaled: https://sentry.slack.com/archives/C04M5GE2R5M/p1718885290037649

RulaKhaled: It looks like we need to create a new github app to verify this

adrian-codecov (approved):

tests/integration/test_github.py
```diff
@@ -169,6 +169,19 @@ async def test_get_authenticated_user(self, codecov_vcr):
...
         }
 
+    @pytest.mark.asyncio
+    async def test_get_authenticated_user_with_public_emails(self, codecov_vcr):
+        code = "71367b9f258ca9a60c44"
+
+        handler = Github(
+            oauth_consumer_token=dict(
+                key="Iv23liSqj8DAO20A3KLA",
+                secret="a6a6397fffea369e54495c88ca469d988ea4ccd2",
```
adrian-codecov, line 179: Quick thing though, are these the legitimate keys/secrets? If so, now that you have the cassette, you can get rid of these/fake them so that no one knows about the secrets themselves. If they are, and you wanna go the extra mile, try rebasing the commit where these were introduced so there's no track of these committed in gh

PR #247, opened 2024-06-12
feat: add repoid fields to the testinstance

joseph-sentry:
Also removes the flaky_status column which is not being used anymore and adds some indexes on the test instance table


shared/django_apps/reports/models.py
```diff
@@ -257,6 +257,9 @@ class Test(models.Model):
     # for example: the same test being run on windows vs. mac
     flags_hash = models.TextField()
 
+    failure_rate = models.FloatField(null=True)
+    commits_where_fail = ArrayField(models.TextField(), null=True)
```
giovanni-guidini, line 261: I hope the application code will enforce a max length on this. We have had issues in the past with array / json objects in the database that grow without a limit.

giovanni-guidini (dismissed):

shared/django_apps/reports/models.py
joseph-sentry: This is more of a rollup field and  will be capped at commits that failed in the last 30 days, so it shouldn't grow unbounded forever.

shared/django_apps/reports/migrations/0021_remove_testinstance_flaky_status_and_more.py
```diff
@@ -0,0 +1,77 @@
...
+"""
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("reports", "0020_alter_reportleveltotals_coverage_and_more"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
adrian-codecov, line 42: Is there logic using this column or has that been deleted prior to this? 

Also wouldn't this be a risky change too since it's ALTERing, therefore locking the table, and going through every row to do so?

adrian-codecov (approved):

PR #249, opened 2024-06-14
Allow filtering of bundle's assets

JerrySentry:
Update a bundle report's `assets` and `total_size` that allows the follow 3 filter options:
- the asset has at least one chunk whose `entry` value is the request boolean
- the asset has at least one chunk whose `initial` value is the request boolean
- the asset belongs to one of the request `asset_type`s

This filtering applies a intersection search, eg if all 3 are selected, all 3 clauses must be true.

closes https://github.com/codecov/engineering-team/issues/1924
closes https://github.com/codecov/engineering-team/issues/1897

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/report.py
+++ b/shared/bundle_analysis/report.py
@@ -2,13 +2,26 @@ import json
 import os
 import sqlite3
 import tempfile
-from typing import Any, Dict, Iterator, Optional, Set, Tuple
+from typing import Any, Dict, Iterator, List, Optional, Set, Tuple
 
 from sqlalchemy import text
 from sqlalchemy.exc import OperationalError
+from sqlalchemy.orm.query import Query
 from sqlalchemy.sql import func
 
-from shared.bundle_analysis import models
+from shared.bundle_analysis.models import (
+    SCHEMA,
+    SCHEMA_VERSION,
+    Asset,
+    AssetType,
+    Bundle,
+    Chunk,
+    Metadata,
+    MetadataKey,
+    Module,
+    Session,
+    get_db_session,
+)
 from shared.bundle_analysis.parser import Parser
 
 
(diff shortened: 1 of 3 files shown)


shared/bundle_analysis/report.py
```diff
@@ -76,40 +89,79 @@ class BundleReport:
...
     """
 
-    def __init__(self, db_path: str, bundle: models.Bundle):
+    def __init__(self, db_path: str, bundle: Bundle):
         self.db_path = db_path
         self.bundle = bundle
 
     @property
     def name(self):
         return self.bundle.name
```
JerrySentry, line 99: This part is the only real change, everything else is import pathing change.
Basically adding to some sqlalchemy query conditions for the 3 filter options.

tests/samples/sample_bundle_stats.json
```diff
@@ -56,7 +56,7 @@
       "id": "index",
       "uniqueId": "2-index",
       "entry": true,
-      "initial": false,
+      "initial": true,
```
adrian-codecov, line 59: Does being initial affect the filtering?

tests/unit/bundle_analysis/test_bundle_analysis.py
```diff
@@ -96,6 +96,45 @@ def test_create_bundle_report():
...
+        }
+
+        bundle_reports = list(report.bundle_reports())
+        assert len(bundle_reports) == 1
+
+        bundle_report = report.bundle_report("invalid")
+        assert bundle_report is None
+        bundle_report = report.bundle_report("sample")
+
+        asset_reports = list(bundle_report.asset_reports(
```
adrian-codecov, line 116: [Optional but encouraged]: could we fetch the non-filtered assets, then the filtered ones (like you do here) to see this filtering the list? Even a "len = 4" or something suffices

adrian-codecov (dismissed):

tests/samples/sample_bundle_stats.json
JerrySentry: Yes, its one of the filtering options

adrian-codecov (approved):

PR #250, opened 2024-06-17
dev: Add perf rules for ruff on shared

ajay-sentry:
### Purpose/Motivation

Part of the lint enhancement epic, this PR aims to add the PerfLint PERF rules.
These can be found here:
- https://docs.astral.sh/ruff/rules/#perflint-perf

All of the fixes were either doing list comprehensions where possible, or using .values() or .keys() when we only needed one or the other. The latter is definitely a micro-optimization, but it improves the clarity of the code as well.

Resources:
- Why list comprehensions are much faster in python 3.12: https://peps.python.org/pep-0709/

### Links to relevant tickets

Closes https://github.com/codecov/engineering-team/issues/1931

### What does this PR do?

This PR adds the rules listed above and fixes any errors that stemmed from them.

### Legal Boilerplate


adrian-codecov (approved):
Pretty sweet to see how this reasons about the change

PR #252, opened 2024-06-18
Add Account, OktaSettings, StripeBilling, InvoiceBilling

nora-codecov:
New models for `Account` project, bringing you the "enterprise experience".

My piece (billing side) write up [here](https://www.notion.so/sentry/Codecov-Billing-67bc332086cb45babdc0da513bdfb566?pvs=4)
@michelletran-codecov 's piece (auth side) write up [here](https://www.notion.so/sentry/Account-Modelling-27888fb38b06483c963e4b15c1889625?pvs=4)

https://github.com/codecov/engineering-team/issues/1740

PR #254, opened 2024-06-19
Set up shared for processing flakes

joseph-sentry:
- Adds reduced error and flake models
- adds boilerplate for using process flakes task
- adds flake detection config option
- adds Test, TestInstance, Flake testing factories

Depends on: https://github.com/codecov/shared/pull/247


shared/django_apps/reports/tests/factories.py
```diff
@@ -103,3 +104,31 @@ class Meta:
...
     )
+
+
+class TestFactory(DjangoModelFactory):
+    class Meta:
+        model = Test
+    
+    repository = factory.SubFactory(RepositoryFactory)
+    id = factory.Sequence(lambda n: f"test_{n}")
+
```
michelletran-codecov, line 115: It looks like the formatter wasn't run. More context: https://github.com/codecov/shared/pull/252#discussion_r1646379852 but please run the formatter manually.

michelletran-codecov (commented):
Generally LGTM. Just one lint comment.

shared/django_apps/reports/migrations/0022_reducederror_testinstance_reduced_error_flake.py
```diff
@@ -0,0 +1,121 @@
...
+                on_delete=django.db.models.deletion.CASCADE,
+                related_name="testinstances",
+                to="reports.reducederror",
+            ),
+        ),
+        migrations.CreateModel(
+            name="Flake",
+            fields=[
+                ("id", models.BigAutoField(primary_key=True, serialize=False)),
+                ("external_id", models.UUIDField(default=uuid.uuid4, editable=False)),
```
michelletran-codecov, line 74: It might be good to sync up and not use `external_id` if we don't need it. (might require model in https://github.com/codecov/shared/pull/252/files#diff-9408bf4ca99f8b12accac14065013a0a52c56db0ba52e5c95729df34a60b1901R16).

michelletran-codecov (approved):
LGTM!

PR #255, opened 2024-06-19
Fix duplicate check in metrics service

rohitvinnakota-codecov:
Also check that the payload is the same before skipping the write. Current tests should catch this logic already. 

### Legal Boilerplate

### DIFF
--- a/shared/django_apps/codecov_metrics/service/codecov_metrics.py
+++ b/shared/django_apps/codecov_metrics/service/codecov_metrics.py
@@ -24,7 +24,7 @@ class UserOnboardingMetricsService:
             return
 
         if not UserOnboardingLifeCycleMetrics.objects.filter(
-            org_id=org_id, event=event
+            org_id=org_id, event=event, additional_data=payload
         ).exists():
             metric = UserOnboardingLifeCycleMetrics(
                 org_id=org_id,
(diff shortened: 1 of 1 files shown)


shared/django_apps/codecov_metrics/service/codecov_metrics.py
```diff
@@ -24,7 +24,7 @@ def create_user_onboarding_metric(org_id: int, event: str, payload: Dict[str, An
             return
 
         if not UserOnboardingLifeCycleMetrics.objects.filter(
-            org_id=org_id, event=event
+            org_id=org_id, event=event, additional_data=payload
```
JerrySentry, line 27: For this entire part I think you can just use `get_or_create`

```
UserOnboardingLifeCycleMetrics.objects.get_or_create(
    org_id=org_id,
    event=event,
    additional_data=payload,
)
```

JerrySentry (approved):

shared/django_apps/codecov_metrics/service/codecov_metrics.py
rohitvinnakota-codecov: Oh cool, ty

JerrySentry (approved):

PR #259, opened 2024-06-21
feat: extend configuration for `comment.required_changes`

giovanni-guidini:
We are extending the configuration for required_changes.
ticket: https://github.com/codecov/engineering-team/issues/1966

I might have overengineered the different conditions, but being able to combine in both AND and OR
operations is somewhat tricky. In the end it's a satisfaction problem in that all the elements in the
conditions list need to be satisfied for the comment to occur, and the elements themselves express
and OR grouping.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,7 @@ setup(
         "cachetools",
         "django-better-admin-arrayfield",
         "django-postgres-extra>=2.0.8",
+        "pyparsing",
         # API Deps
         "django-prometheus",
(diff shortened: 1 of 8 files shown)


joseph-sentry (approved):

giovanni-guidini:
I'll hold off merging this until https://github.com/codecov/worker/pull/520 is approved.
These changes are too significant for the shared code to not be paired up with the worker code in tandem

PR #263, opened 2024-06-25
risky migration part 1

nora-codecov:
part 1 - migration file only


shared/django_apps/codecov_metrics/service/codecov_metrics.py
```diff
@@ -1,8 +1,6 @@
 import logging
 from typing import Any, Dict
 
-from django.utils import timezone
-
```
nora-codecov, line 5: ruff formatting

michelletran-codecov (approved):
LGTM!

PR #265, opened 2024-06-26
[vuln] Bump urllib3

rohitvinnakota-codecov:
Patch urllib3

This closes https://github.com/codecov/internal-issues/issues/551

### Legal Boilerplate


setup.py
```diff
@@ -42,7 +42,7 @@
         "typing_extensions",
         "google-auth>=2.21.0",
         "google-cloud-pubsub>=2.13.6",
-        "urllib3>=1.25.4,<1.27",
```
giovanni-guidini, line 45: DO you know why the requirement was <1.27? Just curious

giovanni-guidini (approved):

setup.py
rohitvinnakota-codecov: I vaguely remember seeing a vuln where v 1.27 - v2.2 were affected, but I can't find it now

I set it `==1.26.19` just to be safe

PR #266, opened 2024-06-26
BA: Handle multiple file versions + introduce gzip size

JerrySentry:
Create a mechanism to handle parsing of different stats file versions, depending on the version value in the JSON file it will choose the appropriate parser logic. The difference between V1 and V2 is that V2 introduces gzip sizes for assets in the JSON file. In V2 parser it will read and store that value, whereas in V1 parser it will just use the old not-smart logic by assuming the gzip size is 1/1000th of the uncompressed size; beyond that V1 and V2 is identical in logic. The gzip size is then stored into the bundle report (ie SQLite DB).

I duplicated the code instead of modularizing it because having a bunch of version specific logic littered within one parser seems more unwieldy over time when there is many versions. Also with the proposed setup it helps with deprecating old ones where we can simply just delete old parsers and older versions' parsers shouldn't be touched most of the time.

closes https://github.com/codecov/engineering-team/issues/1926
closes https://github.com/codecov/engineering-team/issues/1927

### Legal Boilerplate


shared/bundle_analysis/parsers/v1.py
```diff
@@ -0,0 +1,371 @@
...
+    assets_chunks,
+    chunks_modules,
+)
+from shared.bundle_analysis.utils import get_extension
+
+log = logging.getLogger(__name__)
+
+
+"""
+Version 1 Schema
```
giovanni-guidini, line 26: This schema is what the bundler plugin uploads to storage?

JerrySentry: yes for version 1, but some plugins can upload version 2 which is a different schema (see v2.py)

shared/bundle_analysis/parser.py
```diff
@@ -1,336 +1,28 @@
...
-                    for unique_external_id in chunk_unique_external_ids
-                ]
-            )
-        if inserts:
-            self.db_session.execute(chunks_modules.insert(), inserts)
+    def __new__(cls, path: str, db_session: DbSession) -> Optional[object]:
+        with open(path, "rb") as f:
+            for event in ijson.parse(f):
+                prefix, _, value = event
+                if prefix == "version":
```
giovanni-guidini, line 27: [nit] raise exception if there's no "version" in the JSON

adrian-codecov (approved):

PR #267, opened 2024-06-26
Add bots related logic from worker to shared

adrian-codecov:
This PR moves prewritten logic from Worker to Shared. Most of it is copy paste, but I will point out where there are differences.

The intention for this move is to reuse the `get_adapter_auth_information` function between API + Worker to better surface rate limits. This fn is good at determining which token an owner uses, which will help determine other helper functions in here https://github.com/codecov/engineering-team/issues/1977 determine if said token is rate limited. 

This work also intends to replace the existing logic in worker and become the unique source of truth - work to move this is planned here https://github.com/codecov/engineering-team/issues/1969.

### Legal Boilerplate


shared/bots/__init__.py
```diff
@@ -0,0 +1,65 @@
+import logging
```
adrian-codecov, line 1: Copy paste + adjusting imports - no change

shared/bots/exceptions.py
```diff
@@ -0,0 +1,16 @@
+class RequestedGithubAppNotFound(Exception):
```
adrian-codecov, line 1: New helper file for exceptions

shared/bots/github_apps.py
```diff
@@ -0,0 +1,278 @@
+import logging
```
adrian-codecov, line 1: Copy paste + adjusting imports - minimal change below

shared/bots/github_apps.py
```diff
@@ -0,0 +1,278 @@
...
+        keys.pop(key_idx)
+        weights.pop(key_idx)
+        selections += 1
+    if installation_name != GITHUB_APP_INSTALLATION_DEFAULT_NAME:
+        # Add the default app as the last fallback if the owner is using a different app for the task
+        default_apps = filter(
+            lambda obj: _can_use_this_app(
+                obj, GITHUB_APP_INSTALLATION_DEFAULT_NAME, repository
+            ),
+            owner.github_app_installations.all(),
```
adrian-codecov, line 107: Added the `.all()` as django fetches these as RelatedManagers, so doing the .all() turns this into an iterable queryset. This happens in two other places in this file

shared/bots/helpers.py
```diff
@@ -0,0 +1,40 @@
+from typing import Any, Dict
```
adrian-codecov, line 1: Copy paste + adjusting imports - no change

shared/bots/repo_bots.py
```diff
@@ -0,0 +1,74 @@
...
+
+def get_repo_appropriate_bot(repo: Repository) -> Owner:
+    if repo.bot is not None and repo.bot.oauth_token is not None:
+        log.info(
+            "Repo has specific bot",
+            extra=dict(repoid=repo.repoid, botid=repo.bot.ownerid),
+        )
+        return repo.bot
+    try:
+        return get_owner_or_appropriate_bot(repo.author)
```
adrian-codecov, line 38: Changed from repo.owner to repo.author. Sqlalchemy has an `owner` foreign key to an Owner while django calls it `author`

shared/bots/repo_bots.py
```diff
@@ -0,0 +1,74 @@
+import logging
```
adrian-codecov, line 1: Copy paste + adjusting imports - minimal changes below

shared/bots/owner_bots.py
```diff
@@ -0,0 +1,38 @@
+import logging
```
adrian-codecov, line 1: Copy paste + adjusting imports - no change

shared/bots/public_bots.py
```diff
@@ -0,0 +1,78 @@
+import logging
```
adrian-codecov, line 1: Copy paste + adjusting imports - no change

shared/bots/types.py
```diff
@@ -0,0 +1,34 @@
+from typing import Dict, List, Optional, Tuple, TypedDict
```
adrian-codecov, line 1: Copy paste + adjusting imports - no change

shared/django_apps/codecov_auth/migrations/0057_account_stripebilling_oktasettings_invoicebilling_and_more.py
```diff
@@ -1,8 +1,8 @@
 # Generated by Django 4.2.13 on 2024-06-18 00:16
 
+import django.db.models.deletion
```
adrian-codecov, line 3: Linter

shared/environment/environment.py
```diff
@@ -0,0 +1,65 @@
+import os
```
adrian-codecov, line 1: New file that treats with environment stuff, needed for one of the helpers

tests/unit/bots/test_bots.py
```diff
@@ -0,0 +1,628 @@
+import datetime
```
adrian-codecov, line 1: Test logic itself is the same, changed some things to make it work:

- changed all from sqlalchemy models to django models
- got rid of `dbsession`
- adjusted related manager calls to be .all() (and wrapped them in `list` when asserted against a list). This was for repo.github_app_installations and repo.author
- added django_db decorator
- 
-

shared/bots/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+    Service,
+)
+from shared.django_apps.core.models import Repository
+from shared.typings.torngit import GithubInstallationInfo
+
+log = logging.getLogger(__name__)
+
+
+def get_adapter_auth_information(
+    owner: Owner,
```
giovanni-guidini, line 23: If this is required to be a Django model, will this code be able to be used by the worker?

If not (I suspect not at this time) is there a plan to move the worker to use this code? (instead of having this all repeated in the worker as well)

shared/bots/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+    repository: Optional[Repository] = None,
+    *,
+    ignore_installations: bool = False,
+    installation_name_to_use: str | None = GITHUB_APP_INSTALLATION_DEFAULT_NAME,
+) -> AdapterAuthInformation:
+    """Gets all the auth information needed to send requests to the provider"""
```
giovanni-guidini, line 29: I think there should be some comment saying that this is used mostly for the worker for this purpose. The api is supposed to use the person-making-the-request's oauth token so they see only what they are allowed to see.

giovanni-guidini (commented):
some philosophical question/comment

shared/bots/__init__.py
adrian-codecov: The main differences are 
a) repo.owner -> repo.author
b) owner.github_app_installations -> owner.github_app_installations.all()

For a), we could add the author relationship and it should be fine, although I'm not sure if sqlalchemy would complain to have 2 keys with the same relationship. But b) is a django specific workaround so that one would be a bit harder to do. 

We could make it work by passing the django Owner/Repository to this function every time it's used, the DB transactions could get a bit weird, but this logic doesn't seem to persist anything, it's just gets, so it could work

giovanni-guidini: If there is the _intention_ of using only the Django models (including in the worker) for the medium-future I think it's fine to force the interface using Django models.

Otherwise we might think about establishing a common interface (through getter functions) in both models.

shared/bots/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+def get_adapter_auth_information(
+    owner: Owner,
+    repository: Optional[Repository] = None,
+    *,
+    ignore_installations: bool = False,
+    installation_name_to_use: str | None = GITHUB_APP_INSTALLATION_DEFAULT_NAME,
+) -> AdapterAuthInformation:
+    """
+    Gets all the auth information needed to send requests to the provider.
+    This is mostly used by worker and api should use the token provided by the person making the request
```
giovanni-guidini, line 31: [nit] Suggestion for comment (should have suggested earlier, my bad)
```python
    Gets all the auth information needed to send requests to the provider.
    This logic is used by the worker to get the data needed to create a torngit.BaseAdapter
    :warning: Api should use the `owner.oauth_token` of the user making the request
```

giovanni-guidini (approved):
~I personally feel that this will just be duplicating logic for no reason (as the worker can't directly be moved to this logic and api shouldn't use it). It will make it harder to maintain.~

But I know there's a reason that is not directly related with how it's used in the worker. So I'll approve because of that.

I also feel that we should have a plan to move worker to this logic and remove the one on the worker repo.

[update] https://github.com/codecov/engineering-team/issues/1969 we have a plan to move worker to this version of the code 🎉

shared/bots/__init__.py
adrian-codecov: Np! Thanks for the propmpt

PR #270, opened 2024-06-30
fix: include region when initializing storage

imnotjames:
the `api_archive.storage` module isn't pulling in the region as expected which causes the region to be ignored in some cases

this adds the region back similar to the `codecov-api` repo

### Legal Boilerplate


imnotjames:
not sure why this was copied but it means the same changes need to be applied here, too, and they weren't

PR #277, opened 2024-07-04
Adjust bots fns to convert incoming owner + repository models into django models

adrian-codecov:
We're in the process of moving bots related code from worker to shared, see here https://github.com/codecov/shared/pull/267. When doing so, we have to move the sqlalchemy logic to be django friendly as worker is still using sqlalchemy. My first thought was to supply the django model from worker, but I then came up with this implementation, where create proxy classes to deal with the SQLAlchemy case and Django separately, adhering this logic to both django + sqlalchemy for now

Changes also add an integration test I didn't see in the first iteration - we're only moving the preexisting test from worker to shared + creating a cassete for it.

### Legal Boilerplate


giovanni-guidini (commented):
I think that's an interesting approach, it's simple but seems very effective and ergonomic.

It's a little sad that this will inevitably slow us down a little :E but nothing we can do about it until the migration is complete. Shouldn't be too bad, right? Just getting an owner. DB cache is going to help us as well.

matt-codecov (dismissed):
i think `orms/owner_helper.py` and `orms/repository_helper.py` are worth directly writing tests for, but per our chat, maybe those tests should live in Worker (https://github.com/codecov/worker/pull/555) where we can actually create both types of models

without testing those helpers directly, all existing tests around `get_github_integration_token` or `get_adapter_auth_information` are only covering the Django path and the SQLAlchemy path could be broken

nit: the `self` parameter in `_is_django_model()`, `get_github_app_installations()`, and `get_repo_owner()` is unnecessary, and having it means you have to construct an instance of a content-less class just to call those functions. you could get rid of `self` and make them static methods

```
def is_django_model(object):
    return isinstance(object, models.Model)

class DjangoSQLAlchemyOwnerWrapper:
    @staticmethod
    def get_github_app_installations(owner):
        is_django_model(owner):
            ...
        else:
            ...
            
print(DjangoSQLAlchemyOwnerWrapper.get_github_app_installations(owner))
```

technically this way is more efficient, but mostly i think it's a little tidier

matt-codecov (approved):

PR #279, opened 2024-07-05
Sync Account Users when Activating/Deactivating Users

michelletran-codecov:
This adds the feature to add and remove users from an account if the organization that is connected to the account is activating and deactivating users. We want to make sure that users who are deactivated are only removed when they are not associated with _any_ other organization that the account is associated with.

ref: https://github.com/codecov/engineering-team/issues/1993

### Legal Boilerplate

### DIFF
--- a/shared/django_apps/codecov_auth/models.py
+++ b/shared/django_apps/codecov_auth/models.py
@@ -5,10 +5,12 @@ import uuid
 from dataclasses import asdict
 from datetime import datetime
 from hashlib import md5
+from typing import Self
 
 from django.contrib.postgres.fields import ArrayField, CITextField
 from django.contrib.sessions.models import Session as DjangoSession
 from django.db import models
+from django.db.models.fields import AutoField
 from django.db.models.manager import BaseManager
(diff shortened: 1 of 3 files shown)


shared/django_apps/codecov_auth/models.py
```diff
@@ -149,6 +151,37 @@ def __str__(self):
...
+        if not user:
+            user = User(name=user.name, email=user.email)
+        self.activate_user_onto_account(user)
+
+    def deactivate_owner_user_from_account(self, owner_user: "Owner") -> None:
+        if owner_user.user is None:
+            return
+
+        organizations_in_account: list[Owner] = self.organizations.all()
+        all_users_for_account: list[AutoField] = [
```
joseph-sentry, line 169: [nit] this probably doesn't matter for most cases, but since we're checking for membership of this list later and not iterating then I think if AutoField is hashable a set might be a better data structure to use here

joseph-sentry (approved):
lgtm

joseph-sentry (approved):

PR #280, opened 2024-07-08
feat: process merge_commit_sha in get_pull_request

joseph-sentry:
We want to use the merge commit sha to detect
whether a pull was done via squash or merge commit


shared/torngit/bitbucket.py
```diff
@@ -509,6 +523,7 @@ async def get_pull_request(self, pullid, token=None):
             title=res["title"],
             id=str(pullid),
             number=str(pullid),
+            merge_commit_sha=merge_commit["hash"] if merge_commit else None,
```
giovanni-guidini, line 526: Is this hash different from the PR details one?

giovanni-guidini (commented):
Just for my own understanding (I know it's a somewhat stupid question 😅), if the PR has `merge_commit_sha` set we know that it was MERGED (not squashed nor rebased) ?

shared/torngit/bitbucket.py
joseph-sentry: no that would be more convenient but basically if the merge commit sha has multiple parents then it's via merge, else it's via squash

giovanni-guidini: In line 497 you get the `merge_commit` using an api call. Part of the input for that call is `res["merge_commit"]["hash"]`.
And here (line 525) you use `merge_commit["hash"]`.

The question is if `res["merge_commit"]["hash"] == merge_commit["hash"]`

giovanni-guidini (approved):
I updated the question, but depending on the answer lgtm

shared/torngit/bitbucket.py
joseph-sentry: so they wouldn't be equal because the hash that we get in the PR object is truncated, but I think that you're right that we can use it anyways, and skip the get_commit in this function, because we can use that sha to call get_commit anyways which is what we need the `merge_commit_sha` for

giovanni-guidini (approved):

PR #283, opened 2024-07-10
Support for bundle analysis caching pt1

JerrySentry:
An assortment of changes to enable bundle analysis caching:
- Add a table (in postgres) to configure whether a bundle should be cached or not. Also adds a upsert function for the table to be used by the BA processor.
  - closes https://github.com/codecov/engineering-team/issues/2034
- Add a new column for bundles table (in sqlite) to identify whether the bundle is a cached bundle or not. Also adds a update function to set the values for the bundles in the `BundleAnalysisReport` class.
  - closes https://github.com/codecov/engineering-team/issues/2038
- Add a couple of fetching functions to return if a bundle is cached and if a bundle analysis report has any cached bundles.
  - closes https://github.com/codecov/engineering-team/issues/2039
- Add ability to delete a bundle by its name. To be used by BA processor when we no longer want to cache a bundle
  - closes https://github.com/codecov/engineering-team/issues/2041

### Legal Boilerplate


shared/django_apps/bundle_analysis_app/models.py
```diff
@@ -0,0 +1,22 @@
+from django.db import models
+
+# Added to avoid 'doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS' error
+# Needs to be called the same as the API app
```
adrian-codecov, line 4: Don't think we need this comment, makes it sound like we had a BA app in API before but this only applies to django apps we had in API that then moved to Shared

shared/django_apps/dummy_settings.py
```diff
@@ -28,6 +28,7 @@
     "shared.django_apps.codecov_auth",
     "shared.django_apps.core",
     "shared.django_apps.reports",
+    "shared.django_apps.bundle_analysis_app",
```
adrian-codecov, line 31: Any reason why we don't call this "bundle_analysis"?

adrian-codecov (approved):

shared/django_apps/dummy_settings.py
JerrySentry: Was getting an error in API repo when using "bundle_analysis", think Django is getting it confused with some other directory in the repo.

adrian-codecov (approved):

PR #284, opened 2024-07-10
Checks for Adding User to Account

michelletran-codecov:
* Adds some methods to check that a user can be added to an account. It replicates the logic in owners. 
* Adds task boilerplate for the activate account user task

This PR blocks: https://github.com/codecov/worker/pull/556
ticket ref: https://github.com/codecov/engineering-team/issues/1993

### Legal Boilerplate

### DIFF
--- a/shared/django_apps/codecov_auth/models.py
+++ b/shared/django_apps/codecov_auth/models.py
@@ -90,6 +90,8 @@ class User(ExportModelOperationsMixin("codecov_auth.user"), BaseCodecovModel):
     REQUIRED_FIELDS = []
     USERNAME_FIELD = "external_id"
 
+    _okta_loggedin_accounts: list["Account"] = []
+
     class Meta:
         db_table = "users"
         app_label = CODECOV_AUTH_APP_LABEL
@@ -108,6 +110,14 @@ class User(ExportModelOperationsMixin("codecov_auth.user"), BaseCodecovModel):
     def is_authenticated(self):
         # Required to implement django's user-model interface
         return True
+
+    @property
+    def is_github_student(self) -> bool:
+        try:
+            github_owner: Owner = self.owners.get(service=Service.GITHUB)
+        except Owner.DoesNotExist:
+            return False
+        return github_owner.student
 
     def has_perm(self, perm, obj=None):
         # Required to implement django's user-model interface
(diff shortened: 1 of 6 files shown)


shared/django_apps/codecov_auth/models.py
```diff
@@ -151,6 +159,24 @@ def __str__(self):
         str_representation_of_is_active = "Active" if self.is_active else "Inactive"
         return f"{str_representation_of_is_active} Account: {self.name}"
 
+    @property
+    def activated_user_count(self) -> int:
+        # Activated users include all _non_ students. So we'll need to get all the
+        # students and filter them out.
+        return self.users.filter(owners__student=False).count()
+
+    def can_activate_user(self, user: User | None) -> bool:
```
ajay-sentry, line 168: just curious, when would user be None?

ajay-sentry: ahhh okay gotcha. In the event user is None would we return false? I think now it will fall back to the bottom logic checking seats

Was thinking something like 

`
if not user or user in self.users.all():
    return False

if user.is_github_student:
    return True
`

michelletran-codecov: > In the event user is None would we return false?

Nope! The only reason to pass in the user is to check that they've already been added. If there was no user object associated, then it implicitly means that they haven't already been added to the account (and hence, we should then see if we can add them via the seat count check).

I guess having `User` be None is not super clear that it means that the owner is not already added to the account. I've considered maybe just removing the user-account link check but I felt like having it here was more convenient and more elegant in the client code (i.e specifically [here](https://github.com/codecov/worker/pull/556/files#diff-707fe54f561ef674e783cc52d7d93cc0e5760e9cb744b11d73776926e68d39b8R42)). Not sure if you have a strong opinion on one way or the other.

ajay-sentry: HMmmMmmmm I don't have too much of a preference tbh

Maybe a small comment for when user could be None should be sufficient, especially since that seems like a legacy concept we shouldn't run into much / at all over time. 

Even that seems a little overkill though 😂

michelletran-codecov: Hmm... thinking about it a bit more, I think that we should make this optional. We should be able to call `can_activate_user()` with no params (as it shouldn't matter if the function has a user or not... we're concerned about seat count). I'll also add a comment.

ajay-sentry: That makes sense to me! I think it's a lot more clear that way too

ajay-sentry (approved):
looks good!

PR #287, opened 2024-07-15
feat: bundle threshold coercion

giovanni-guidini:
ticket: https://github.com/codecov/engineering-team/issues/2004

The bundle threshold needs to also accept percentages.
To do that we are going to re-use the percentage schema field and the
byte size schema field, and just have a wrapper around them.

A new return type has been created to make the difference between
absolute and percentage values more explicit (although technically
the values are different - float vs int - it will be easier to distinguish
if we have the type explicitly

### DIFF
--- a/shared/validation/helpers.py
+++ b/shared/validation/helpers.py
@@ -7,6 +7,7 @@ from typing import Any, List
 import pyparsing as pp
 
 from shared.validation.types import (
+    BundleThreshold,
     CoverageCommentRequiredChanges,
     CoverageCommentRequiredChangesANDGroup,
     CoverageCommentRequiredChangesORGroup,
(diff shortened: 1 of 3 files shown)


Swatinem (commented):
code lgtm, though its unclear from the tests whats supposed to happen with other edge cases, like:
- are negative numbers supported at all? for either absolute or percentages?
- are percentages > 100% supported?

Might be good to cover those cases either as passing or failing.

giovanni-guidini:
Those are good points, @Swatinem. Thanks for raising that.

Negative numbers shouldn't match the regex, so they would fail I think (which is what we expect), but the percentage more than 100% shouldn't be accepted, I guess...

In any case I'll add unit tests to make those expectations clearer.

Swatinem (approved):

PR #292, opened 2024-07-18
feat: add migration for fixing upload_id in user_measurements

joseph-sentry:
fixes: https://github.com/codecov/internal-issues/issues/628


shared/django_apps/user_measurements/migrations/0003_auto_20240718_1339.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("user_measurements", "0002_remove_usermeasurement_i_owner_and_more"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            "ALTER TABLE user_measurements ALTER COLUMN upload_id TYPE int8;",
```
adrian-codecov, line 13: Thoughts on changing the field from `IntegerField` to `BigIntegerField` instead? I'm assuming bigint is int8, cause otherwise the db type and the model definition would be out of sync 👎

adrian-codecov (approved):

PR #296, opened 2024-07-19
[FIx] Update is admin API call for bitbucket

rohitvinnakota-codecov:
Screenshot 2024-07-19 at 11 43 38 AM

The `uuid` does not support filtering the bitbucket api. Here we filter manually after we retrieve the result to check if the user is an admin. 

### Legal Boilerplate

### DIFF
--- a/shared/torngit/bitbucket.py
+++ b/shared/torngit/bitbucket.py
@@ -212,21 +212,17 @@ class Bitbucket(TorngitBaseAdapter):
 
     async def get_is_admin(self, user, token=None):
         user_uuid = "{" + user["service_id"] + "}"
-        query = f'workspace.uuid="{{{self.data["owner"]["service_id"]}}}" AND permission="owner" AND user.uuid="{user_uuid}"'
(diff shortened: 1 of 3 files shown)


shared/torngit/bitbucket.py
```diff
@@ -212,20 +212,19 @@ async def delete_webhook(self, hookid, token=None):
 
     async def get_is_admin(self, user, token=None):
         user_uuid = "{" + user["service_id"] + "}"
-        query = f'workspace.uuid="{{{self.data["owner"]["service_id"]}}}" AND permission="owner" AND user.uuid="{user_uuid}"'
+        workspace_uuid = f'{{{self.data["owner"]["service_id"]}}}'
```
michelletran-codecov, line 215: Nit: thoughts on making these consistent? They both add `{ }` around the variable, so it would be more consistent if both used f strings or concatenated `{ }`.

michelletran-codecov (dismissed):

michelletran-codecov (approved):

Swatinem (approved):

PR #300, opened 2024-07-25
add more model properties to Account

nora-codecov:
Planning to use these fields for the read-only experience in api

### DIFF
--- a/shared/django_apps/codecov_auth/models.py
+++ b/shared/django_apps/codecov_auth/models.py
@@ -168,12 +168,36 @@ class Account(BaseModel):
         return self.users.filter(owners__student=False).count()
 
     @property
+    def activated_student_count(self) -> int:
+        return self.users.filter(owners__student=True).count()
+
+    @property
     def all_user_count(self) -> int:
         return self.users.count()
 
     @property
     def organizations_count(self) -> int:
         return self.organizations.all().count()
+
+    @property
+    def total_seat_count(self) -> int:
+        return self.plan_seat_count + self.free_seat_count
+
+    @property
+    def available_seat_count(self) -> int:
+        count = self.total_seat_count - self.activated_user_count
+        return count if count > 0 else 0
+
+    @property
+    def pretty_plan(self) -> dict | None:
+        """
+        This is how we represent the details of a plan to a user, see plan.constants.py
+        We inject quantity to make plan management easier on api, see PlanSerializer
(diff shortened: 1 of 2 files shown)


shared/django_apps/codecov_auth/models.py
```diff
@@ -175,6 +179,26 @@ def all_user_count(self) -> int:
...
+    @property
+    def available_seat_count(self):
+        count = self.total_seat_count - self.activated_user_count
+        return count if count > 0 else 0
+
+    @property
+    def pretty_plan(self):
+        """
+        This is how we represent the details of a plan to a user, see plan.constants.py
+        We inject quantity to make plan management easier on api, see PlanSerializer
```
giovanni-guidini, line 195: I like the code pointers, very helpful

shared/django_apps/codecov_auth/models.py
```diff
@@ -175,6 +179,26 @@ def all_user_count(self) -> int:
...
+    def total_seat_count(self) -> int:
+        return self.plan_seat_count + self.free_seat_count
+
+    @property
+    def available_seat_count(self):
+        count = self.total_seat_count - self.activated_user_count
+        return count if count > 0 else 0
+
+    @property
+    def pretty_plan(self):
```
giovanni-guidini, line 192: return type annotation (`None` is a possiblity apparently)

shared/django_apps/codecov_auth/models.py
```diff
@@ -167,6 +167,10 @@ def activated_user_count(self) -> int:
         # students and filter them out.
         return self.users.filter(owners__student=False).count()
 
+    @property
```
giovanni-guidini, line 170: Should we consider using [@cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) in some of these that hit the DB? Maybe the DB itself already has a cache that we can fallback on?

giovanni-guidini (dismissed):
LGTM, left some comments

shared/django_apps/codecov_auth/models.py
nora-codecov: we could - it's an interesting idea. I'm hoping that this won't be a heavy/expensive query, I'm going to add it as-is and if it's slow I'll upgrade it to `@cached_property`

michelletran-codecov (approved):

PR #303, opened 2024-07-30
Add fns to rate limit keys

adrian-codecov:
We're making use of some fns to rate limit gh apps + owners via redis. The current code does that herehttps://github.com/codecov/shared/blob/2f846bffec0b8391362fd8d8acd67dd15079c6ed/shared/torngit/github.py#L735. This only works for gh_apps, so we want to extend this functionality to owner records as well. 

This PR will change some of the functions internals as well as move around when they're called to ensure applications are rate limited when the rate limit is hit. 

- Replaces `is_installation_rate_limited` with `determine_if_entity_is_rate_limited` and `gh_app_key_name`
- Replaces `mark_installation_as_rate_limited` with `set_entity_to_rate_limited`
- Adds `determine_entity_redis_key_from_torngit_data`, an alternate function to `determine_entity_redis_key` when you already have the TorngitInstanceData + helper methods that normalize the redis rate limited key names
- Modified the fns in `rate_limits/__init__.py` to add better descriptions, edge cases, error handling, etc'
- Made `_possibly_mark_current_entity_as_rate_limited` use `set_entity_to_rate_limited`. Also moved its call outside `_get_next_fallback_token` to ensure we rate limit the token regardless of another token option.
- Added + adjusted tests

### Legal Boilerplate


shared/rate_limits/__init__.py
```diff
@@ -1,17 +1,52 @@
...
+    current_installation = data.get("installation")
+    current_owner = data.get("owner")
+
+    if current_installation and current_installation.get("installation_id"):
+        return gh_app_key_name(
+            app_id=current_installation.get("app_id"),
+            installation_id=current_installation.get("installation_id"),
+        )
+    elif current_owner and current_owner.get("ownerid"):
+        return owner_key_name(owner_id=current_owner.get("ownerid"))
```
giovanni-guidini, line 45: I think there's an edge case here when the repo has a bot configured.

Notice that the AdapterAuthInfo includes the [token_owner](https://github.com/codecov/shared/blob/6ac0e7a2f571120969ccc0745c84444f45520ac4/shared/bots/__init__.py#L67), currently used for the token refresh functions.
The TorngitInstanceData on the other hand always uses [the repo owner](https://github.com/codecov/worker/blob/ed134ad2d06ab8f3a20738ee53e32762c41abeb2/services/repository.py#L108) as the owner on data.
This means that the token that is being used by the torngit instance might not actually belong to the owner in `data.get("owner")`, leading us to mark the `repo.owner` as rate limited when in fact is `repo.bot.owner` that has been rate limited

shared/torngit/github.py
```diff
@@ -731,10 +729,6 @@ def _get_next_fallback_token(
         if fallback_installations is None or fallback_installations == []:
             # No tokens to fallback on
             return None
-        # ! side effect: mark current token as rate limited
```
giovanni-guidini, line 734: update the function docstring

shared/rate_limits/__init__.py
```diff
@@ -1,17 +1,52 @@
...
+    else:
+        return bot_key_name()
```
giovanni-guidini, line 47: I think there's an edge case here when we use one of the TokenTypeMapping tokens in a particular request.

Sometimes we don't use the "main" token in the Torngit instance, but one of the tokens in TokenTypeMapping. These are shared among guest users for specific requests. If _that_ token is rate limited it doesn't mean necessarily that the github_bot is rate limited

This edge case also affects `determine_entity_redis_key` :E

shared/rate_limits/__init__.py
```diff
@@ -1,17 +1,52 @@
...
+
+def owner_key_name(owner_id: int) -> str:
+    return str(owner_id)
+
+
+def bot_key_name() -> str:
+    return str(get_config("github", "bot", "key"))
+
+
+# TODO: Test this
```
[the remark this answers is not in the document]
adrian-codecov: Naah I tested this, I made the commit to get rid of this line and didn't push it 🙃

shared/torngit/github.py
adrian-codecov: tytyt

shared/rate_limits/__init__.py
adrian-codecov: Mmm so in this case, is would `token_owner` end up being `repo.bot.owner`?

shared/rate_limits/__init__.py
adrian-codecov: Can you help me understand this a bit further? Not sure I fully follow. 

For this and [this](https://github.com/codecov/shared/pull/303#discussion_r1697865195) comment, the goal is to say "this entity (if available) will be rate limited because it's associated token is rate limited". The get_auth_adapter fn in theory does that right, get you who the entity (app/owner) + its token is - I suppose the token itself here doesn't matter as much if it's _always_ going to be the same for the person, but if it rotates + changes, then we'd probably need to adjust the fn to contemplate a combination of app/owner + token

shared/bots/github_apps.py
```diff
@@ -118,13 +119,18 @@ def get_github_app_token(
...
-        installation_info["installation_id"],
-        app_id=installation_info.get("app_id", None),
+        installation_id,
+        app_id=app_id,
         pem_path=installation_info.get("pem_path", None),
     )
-    installation_token = Token(key=github_token)
+    installation_token = Token(
+        key=github_token,
+        entity_name=gh_app_key_name(installation_id=installation_id, app_id=app_id),
```
adrian-codecov, line 132: Please double check I didn't make the wrong assumptions when defining entity_names

shared/torngit/github.py
```diff
@@ -676,47 +678,44 @@ def _parse_response(self, res: Response):
...
-                    extra=dict(
-                        installation_id=installation_id,
-                        app_id=app_id,
-                        rate_limit_duration_seconds=ttl_seconds,
-                    ),
-                )
-                mark_installation_as_rate_limited(
-                    self._redis_connection, installation_id, ttl_seconds, app_id=app_id
-                )
+        entity_key_name = token.get("entity_name")
```
adrian-codecov, line 689: Here's where I'm somewhat unsure what to do when it comes to a token that belongs to the token mappings, cause technically it wouldn't have an entity name

tests/unit/bots/test_bots.py
```diff
@@ -66,7 +66,10 @@ def test_select_owner_info(self):
             owner = self._generate_test_owner(with_bot=False)
             expected = AdapterAuthInformation(
                 token=Token(
-                    key="owner_token", refresh_token="refresh_token", secret=None
+                    key="owner_token",
```
adrian-codecov, line 69: This is probably the most important file to check to ensure the entity_name logic makes sense

shared/rate_limits/__init__.py
```diff
@@ -39,33 +55,49 @@ def determine_entity_redis_key(
...
-        return f"{auth_info.get("selected_installation_info").get("app_id")}_{auth_info.get("selected_installation_info").get("installation_id")}"
-    # How to treat gh app fallbacks?
-    else:
-        return str(auth_info.get("token_owner").ownerid)
+    if (
+        auth_info
+        and auth_info.get("token")
+        and auth_info.get("token").get("entity_name")
+    ):
+        return auth_info.get("token").get("entity_name")
```
giovanni-guidini, line 64: it looks like this function can also return `None`. Please update the return type

shared/rate_limits/__init__.py
```diff
@@ -39,33 +55,49 @@ def determine_entity_redis_key(
...
 def determine_if_entity_is_rate_limited(redis_connection: Redis, key_name: str) -> bool:
     """
     This function will determine if a customer is rate limited. It will
     return true if the record exists, false otherwise.
-    This will be used by API and Worker
+    This will be used by API and Worker and should only be used for github git instances.
     """
-    return True if redis_connection.get(f"rate_limited_entity_{key_name}") else False
+    try:
+        return redis_connection.exists(f"rate_limited_entity_{key_name}")
```
giovanni-guidini, line 73: [nit] it might be a good idea to extract `rate_limited_entity_` into a constant.

```python
RATE_LIMIT_REDIX_KEY_PREFIX="rate_limited_entity_"
...
return redis_connection.exists(f"{RATE_LIMIT_REDIX_KEY_PREFIX}{key_name}")
```

shared/rate_limits/__init__.py
```diff
@@ -39,33 +55,49 @@ def determine_entity_redis_key(
...
-def set_entity_to_rate_limited(redis_connection, key_name: str, ttl_seconds: int):
+def set_entity_to_rate_limited(
+    redis_connection: Redis, key_name: str, ttl_seconds: int
+):
     """
-    This function will set a key to be rate limited. This will be mainly used
+    Marks an entity as rate-limited in Redis. This will be mainly used
     in worker during communication with Github 3rd party services
 
+    @param `key_name` - name of the entity being rate limited. This is found in determine_entity_redis_key
```
giovanni-guidini, line 89: it can also be found in the Token, no?

shared/torngit/github.py
giovanni-guidini: hm... maybe check if there is an entity_key_name before proceeding?
We don't mark those bots as rate limited so... it should be fine right? Otherwise add keys to their tokens

```python
if entity_key_name is None:
  log.warning(
     "Can't mark entity as rate limited because TTL is missing",
  )
  return
```

tests/unit/rate_limits/test_rate_limit.py
```diff
@@ -157,9 +160,63 @@ def test_determine_if_entity_is_rate_limited_false(self):
...
+        mock_redis.set.side_effect = RedisError
+        key_name = "owner_id_123"
+        # This actually asserts that the error is not raised
+        # Despite the call failing
+        set_entity_to_rate_limited(
+            redis_connection=self.redis_connection, key_name=key_name, ttl_seconds=300
+        )
+        mock_redis.set.assert_not_called()
+
+    def test_gh_app_key_name(self):
```
giovanni-guidini, line 203: This test and the next would benefit from `@pytest.mark.parametrize`

tests/unit/rate_limits/test_rate_limit.py
```diff
@@ -157,9 +160,63 @@ def test_determine_if_entity_is_rate_limited_false(self):
...
+        mock_redis.set.assert_not_called()
+
+    def test_set_entity_to_rate_limited_ttl_zero(self):
+        mock_redis = MagicMock(name="fake_redis")
+        mock_redis.set.side_effect = RedisError
+        key_name = "owner_id_123"
+        # This actually asserts that the error is not raised
+        # Despite the call failing
+        set_entity_to_rate_limited(
+            redis_connection=self.redis_connection, key_name=key_name, ttl_seconds=300
```
giovanni-guidini, line 199: I suppose the TTL should be 0 here.
It's weird the test passed then. It might be passing on the wrong expectation

giovanni-guidini (commented):
LGTM, left some comments

shared/rate_limits/__init__.py
adrian-codecov: yeap, good catch, thanks for being so attentive 🙇 slip so easily when doing back and forth reviews 😅

tests/unit/rate_limits/test_rate_limit.py
adrian-codecov: Oh it passed because I accidentally copy pasted a test, changed the title name, and not the implementation. I will fix that

shared/torngit/github.py
```diff
@@ -676,47 +678,49 @@ def _parse_response(self, res: Response):
...
+        if entity_key_name is None:
+            log.warning(
+                "Can't mark entity as rate limited because TTL is missing",
```
giovanni-guidini, line 692: Wrong reason :E

tests/unit/rate_limits/test_rate_limit.py
```diff
@@ -157,9 +160,58 @@ def test_determine_if_entity_is_rate_limited_false(self):
...
+        ],
+    )
+    def test_gh_app_key_name_with_or_without_id(self, app_id, installation_id):
+        assert (
+            gh_app_key_name(
+                app_id=app_id,
+                installation_id=installation_id,
+            )
+            == f"{app_id}_{installation_id}"
+            if app_id
```
giovanni-guidini, line 215: [nit] We try to avoid if/else statements in test cases.

I think the clearer way to do this would be to have an extra parameter `expected`. And then you test simply

```
assert gh_app_key_name(app_id=app_id, installation_id=installation_id) == expected
```

giovanni-guidini (approved):
LGTM, 2 more comments.
(Please fix the warning before merging)

Gguidini (approved):

giovanni-guidini (approved):

giovanni-guidini (approved):

PR #304, opened 2024-07-30
Fix deactivation of users on accounts

michelletran-codecov:
Currently the deactivation code is attempting to remove the incorrect owner reference (the org) when it's supposed to remove the user. This fixes it to remove the correct owner.

### Legal Boilerplate

### DIFF
--- a/shared/django_apps/codecov_auth/models.py
+++ b/shared/django_apps/codecov_auth/models.py
@@ -228,6 +228,9 @@ class Account(BaseModel):
 
     def deactivate_owner_user_from_account(self, owner_user: "Owner") -> None:
         if owner_user.user is None:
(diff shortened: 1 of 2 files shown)


shared/django_apps/codecov_auth/models.py
```diff
@@ -228,6 +228,9 @@ def activate_owner_user_onto_account(self, owner_user: "Owner") -> None:
 
     def deactivate_owner_user_from_account(self, owner_user: "Owner") -> None:
         if owner_user.user is None:
+            log.warning(
```
giovanni-guidini, line 231: good call 👍 
This will be great for debugging

giovanni-guidini (approved):

PR #306, opened 2024-07-30
Adjustments to Accounts flow, plus adjustment to AccountsUsers for QoL improvement

nora-codecov:
gave AccountsUsers a `verbose_name_plural` because seeing `AccountsUserss` was painful.

Adjustments to the Accounts flow with activation and deactivation of users:
- we were doing the creation wrong when an `Owner` didn't have a `User`
- we were counting seats wrong when someone was already an `AccountUser`
- hierarchy was wrong when calculating `can_activate`: if an `OwnerOrg` has an `Account`, ignore what's on the `OwnerOrg`.
- changed variable naming so object types are more clear (we keep tripping ourselves up with the web of relationships between these objects)


michelletran-codecov (approved):
Thanks for the fixes! This LGTM! 🌟

PR #307, opened 2024-07-31
fix: add notify_error to config schema

joseph-sentry:


giovanni-guidini (commented):
Please add a `meta` key with a little description describing what this key does.
Even if it's simple it would be good to have this custom to help the tam in the future that will maintain the user schema, so they don't have to guess what this is supposed to do

see an example [here](https://github.com/codecov/shared/blob/cee7681d68ce0d9f57899d41a6241d0772b07905/shared/validation/user_schema.py#L152)

giovanni-guidini (approved):

PR #316, opened 2024-08-07
Bundle analysis: make gzip size nullable

JerrySentry:
Make a SQLite migration to make it so that gzip size is nullable and when parsing `gzipSize` handles the case that the value can be null.

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/db_migrations.py
+++ b/shared/bundle_analysis/db_migrations.py
@@ -3,6 +3,9 @@ from sqlalchemy.orm import Session
 
 from shared.bundle_analysis.migrations.v001_add_gzip_size import add_gzip_size
 from shared.bundle_analysis.migrations.v002_bundle_is_cached import add_is_cached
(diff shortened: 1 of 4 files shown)


giovanni-guidini (commented):
Don't you have to change [the model](https://github.com/codecov/shared/blob/38cddf96fd55d5e3ebbca2bbcf8689a6da86afe4/shared/bundle_analysis/models.py#L226) too?

I think that's where the default value of `None` should be set, right?...

JerrySentry:
> Don't you have to change [the model](https://github.com/codecov/shared/blob/38cddf96fd55d5e3ebbca2bbcf8689a6da86afe4/shared/bundle_analysis/models.py#L226) too?
> 
> I think that's where the default value of `None` should be set, right?...

@giovanni-guidini Good catch, updated!

giovanni-guidini (approved):

PR #318, opened 2024-08-08
Clean up `requirements` and add `.envrc`

Swatinem:
The `.envrc` file in conjunction with `direnv` will allow to automatically set up and activate a local `.venv` ~which allows running `pytest` and friends locally outside of docker~. This local `.venv` also fixes my local vscode editor completions and type hints, which is a huge plus.

Another change here is that this removes `django-admin`. I don’t see this being used anywhere, and it fails to compile locally, as Python 3.12 removed some deprecated functionality that is required for building `screen` locally, which is being pulled in transitively via `django-admin`. Why this all works in docker though is a mystery to me.

Otherwise it just alphabetically sorts all the various requirements files/lists.

---

Disclaimer: I don’t really consider myself a Python developer, so I’m not entirely sure what I am doing here.

Turns out running a local `pytest` is still not possible, as it seems to expect postgres to be running on the `postgres` hostname, which only exists in docker.


matt-codecov (approved):
i think the worker repo has some tests disabled if run outside of docker, could maybe set that up here. i think probably a better way to improve the test workflow in shared would be to make shared tests runnable inside of umbrella-hat, but as long as the full suite at least runs in CI i'm not opposed to enabling dockerless local test runs for a subset

PR #319, opened 2024-08-08
fix: cant use flags in validator?

giovanni-guidini:
context: https://github.com/codecov/engineering-team/issues/2223

it seems that we get a validation error when trying to coerce the bundle_analysis threshold.
I can only assume it's the use of DJango database from an async context, so removing that.

### DIFF
--- a/shared/validation/validator.py
+++ b/shared/validation/validator.py
@@ -51,11 +51,8 @@ class CodecovYamlValidator(Validator):
         coerced_value: BundleThreshold = BundleSizeThresholdSchemaField().validate(
             value
         )
-        # This change is not forwards compatible, so we need this step in the rollout
-        from shared.rollouts.features import BUNDLE_THRESHOLD_FLAG
-
-        if BUNDLE_THRESHOLD_FLAG.check_value(identifier=1, default=False):
-            return coerced_value
+        # This change is not forwards compatible, so we need to return a number for now
+        # https://github.com/codecov/engineering-team/issues/2087 to clean up later
         return coerced_value.threshold
 
     def _validate_comma_separated_strings(self, constraint, field, value):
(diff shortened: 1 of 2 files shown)


tests/unit/validation/test_validation.py
```diff
@@ -838,7 +838,8 @@ def test_validate_jacoco_partials(self):
                 {
                     "comment": {
                         "require_bundle_changes": True,
-                        "bundle_change_threshold": ("absolute", 1200),
+                        # "bundle_change_threshold": ("absolute", 1200),
```
michelletran-codecov, line 841: Hmm, can we remove the commented code, or link the ticket to clean up here? Commented out code (especially without context) can get confusing/stale pretty quick.

shared/validation/validator.py
```diff
@@ -51,11 +51,8 @@ def _normalize_coerce_bundle_analysis_threshold(self, value):
         coerced_value: BundleThreshold = BundleSizeThresholdSchemaField().validate(
             value
         )
-        # This change is not forwards compatible, so we need this step in the rollout
-        from shared.rollouts.features import BUNDLE_THRESHOLD_FLAG
-
-        if BUNDLE_THRESHOLD_FLAG.check_value(identifier=1, default=False):
```
michelletran-codecov, line 57: I think the `Feature` class has a `check_value_async` https://github.com/codecov/shared/blob/e20db3999ad82d123fa53a0aae01ba2177c94d81/shared/rollouts/__init__.py#L162 . Does that not work here?

michelletran-codecov (commented):
Just a few comments.

shared/validation/validator.py
giovanni-guidini: The function is not async... I'm not sure

michelletran-codecov: > The function is not async.

True. I'm not super familiar with how the worker's async context works, so I just wanted to see if that method in `Feature` would be a quick fix. If it's more complicated, then let's defer the investigation to the ticket.

michelletran-codecov (approved):

PR #322, opened 2024-08-09
fix: if data.installation is None we fail, not anymore

giovanni-guidini:


shared/torngit/github.py
```diff
@@ -2325,11 +2325,9 @@ def loggable_token(self, token) -> str:
             return f"{username}'s token"
         if token is None or token.get("key") is None:
             return "notoken"
-        app_installation = self.data.get("installation", {}).get(
-            "installation_id", None
-        )
-        if app_installation:
-            return f"GitHub_installation_{app_installation}"
+        installation_info = self.data.get("installation", {})
```
Swatinem, line 2328:
```
>>> print({"foo":None}.get("foo", {}))
None
```

Just Python things I guess :-D

Swatinem (approved):

shared/torngit/github.py
giovanni-guidini: sadly not the 1st time I slip on this 😅

PR #324, opened 2024-08-09
Add tracing to Report serialization

Swatinem:


shared/reports/resources.py
```diff
@@ -153,7 +153,6 @@ def lines(self):
                         continue
                 yield ln, line
 
-    @sentry_sdk.trace
```
michelletran-codecov, line 156: What's the reason for removing the trace here? Is this not useful?

michelletran-codecov (approved):
LGTM! Just non-blocking question.

shared/reports/resources.py
Swatinem: The spans on `ReportFile` are generally high-noise, as a `Report` can have hundreds or thousands of files.
I’m trying to avoid tracing things that happen in loops, and I believe its generally a good idea to keep tracing spans bounded to a low number.

`Report.calculate_diff` is also directly called from `Report.apply_diff` which is being traced.

michelletran-codecov: These are all good points! Thanks for the context!

PR #327, opened 2024-08-12
bundle analysis: save custom compare sha to metadata

JerrySentry:
Also have the ingester return the bundle name of the parsed bundle file, this is an optimization for worker to save BA measurements data.

Note: only merge this when the worker PR to match the new ingest function signature is ready to be merged.

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/models.py
+++ b/shared/bundle_analysis/models.py
@@ -112,7 +112,6 @@ class LegacySessionManager:
 
     def __exit__(self, type, value, traceback):
         self.session.close()
-        return True
 
 
 def _use_modern_sqlalchemy_session_manager():
--- a/shared/bundle_analysis/parsers/v2.py
+++ b/shared/bundle_analysis/parsers/v2.py
@@ -92,7 +92,7 @@ class ParserV2:
         self.chunk_list = []
         self.module_list = []
 
-    def parse(self, path: str) -> int:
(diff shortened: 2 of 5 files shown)


shared/bundle_analysis/models.py
```diff
@@ -112,7 +112,6 @@ def __enter__(self):
 
     def __exit__(self, type, value, traceback):
         self.session.close()
-        return True
```
JerrySentry, line 115: Returning True here actually suppresses exceptions being thrown by the DB session context manager

shared/bundle_analysis/parsers/v2.py
```diff
@@ -110,20 +110,6 @@ def parse(self, path: str) -> int:
...
-                    insert_chunks = Chunk.__table__.insert().values(self.chunk_list)
-                    self.db_session.execute(insert_chunks)
-
-                if self.module_list:
-                    insert_modules = Module.__table__.insert().values(self.module_list)
-                    self.db_session.execute(insert_modules)
-
-                self.db_session.flush()
-
                 # Delete old session/asset/chunk/module with the same bundle name if applicable
```
JerrySentry, line 113: Delete the old bundle (and its assets/chunks/modules) first before inserting new ones, this fixes a bug relating to bundle caching.

adrian-codecov (approved):

PR #328, opened 2024-08-14
fix: configure cache for installation tokens

giovanni-guidini:
The function that gets installation tokens is cached so we don't ask for them so often.
However the cache was not being initialized.

SO the increase in requests is probably being interpreted by GitHub as rate limit violation and now we are getting 403s left and right :E

These changes configure the cache (I had to move some redis funcitons) and extends the timeout to 30minutes. THe [docs](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app) say they expire after 1h, so this should be safe to do.

### DIFF
--- a/shared/bots/helpers.py
+++ b/shared/bots/helpers.py
@@ -5,15 +5,20 @@ from shared.bots.exceptions import RepositoryWithoutValidBotError
 from shared.config import get_config
 from shared.github import InvalidInstallationError
 from shared.github import get_github_integration_token as _get_github_integration_token
-from shared.helpers.cache import OurOwnCache
+from shared.helpers.cache import OurOwnCache, RedisBackend
+from shared.helpers.redis import get_redis_connection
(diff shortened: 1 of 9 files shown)


shared/bots/helpers.py
```diff
@@ -5,15 +5,20 @@
...
 cache = OurOwnCache()
+cache.configure(RedisBackend(get_redis_connection()), app="shared.installation_tokens")
+
 log = logging.getLogger(__name__)
 
 
-@cache.cache_function(ttl=480)
+# The integration tokens are valid for 1h
+# We use 30min of that
+@cache.cache_function(ttl=1800)
```
Swatinem, line 21:
```suggestion
# We use 30min of that
@cache.cache_function(ttl=30 * 60)
```

IMO thats a bit more readable :-)

Swatinem (approved):

PR #329, opened 2024-08-15
delete SessionTotalsArray, remove from report

matt-codecov:
https://github.com/codecov/internal-issues/issues/515

we expect the indices in `SessionTotalsArray` to correspond to the session IDs in a `Report`, but the logic for choosing them is different and they can get out-of-sync. when that happens, a ton of orphaned/corrupted data accumulates in `SessionTotalsArray`

from what i've seen, we never actually _use_ `session_totals` anywhere, so instead of rewiring `Report` so we can keep the session IDs in sync, i think we should just delete the field

this needs worker and API PRs ready to go before it can merge. also will need to make `session_totals` optional in the codecov-rs report json parser


tests/unit/reports/test_carryforward.py
```diff
@@ -55,64 +53,8 @@ def sample_report():
...
-                messages=0,
-                sessions=0,
-                complexity=0,
-                complexity_total=0,
-                diff=0,
-            ),
-            None,
-        ]
-    )
+    report._files["file_1.go"]
```
Swatinem, line 56: I think `_files` is just a straight `dict`, so accessing anything on it should be sideeffect free, so you can just remove the two lines completely.

shared/reports/resources.py
```diff
@@ -598,7 +593,17 @@ def build_files(files: dict[str, Any]) -> dict[str, ReportFileSummary]:
...
-            files[filename] = ReportFileSummary(*file_summary)
+            chunks_index = file_summary[0]
+            file_totals = file_summary[1]
+
+            if len(file_summary) > 2:
+                # Old reports have `session_totals` but new reports omit it
+                # Graph `diff_totals` from the back of the list either way
+                diff_totals = file_summary[-1]
+            else:
+                diff_totals = None
```
Swatinem, line 604: Just to double-check:
The format has always been serialized as a 4-element tuple? So there is no backwards compatibility with very ancient reports that have a 3-element `chunks_index, file_totals, session_totals` floating around?

If such backwards compatibility is a concern, I would suggest to keep this as 4-element tuple, but just use an empty value for the `session_totals`.

Swatinem (approved):
lgtm, very happy to see this code being removed.

I added a comment regarding backwards compatibility.
I didn’t see any test changes related to ensuring the "parser" can still deal with inputs that have the legacy 4-element tuple? Most likely because those fixtures weren’t touched in this PR? Just want make sure to double-check these things.

shared/reports/resources.py
matt-codecov: https://github.com/codecov/codecov.io/commit/d0c28c0899a732a47db9882aa73cbb8d580161ce this appears to be the commit where the field was introduced. i did not see any usages of it lol (the `sum_session_totals` SQL function in that same commit works on the `"sessions"` key in the report JSON, not these session totals. and also the sql function is deleted now by a later migration)

you are right about the backcompat, i will update

tests/unit/reports/test_carryforward.py
matt-codecov: oh wow lol i really had tunnel vision deleting those references to not notice this

Swatinem (approved):

PR #330, opened 2024-08-15
bundle analysis: compare base commit with pre commit merge SHA

JerrySentry:
When doing comparisons first check if there is a compare_sha set in the head report, if there is use that commitid to load the base commit report to compare the head to.

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/comparison.py
+++ b/shared/bundle_analysis/comparison.py
@@ -5,6 +5,7 @@ from enum import Enum
 from functools import cached_property
 from typing import Iterator, List, MutableSet, Optional, Tuple
 
+from shared.bundle_analysis.models import MetadataKey
 from shared.bundle_analysis.report import (
     AssetReport,
     BundleAnalysisReport,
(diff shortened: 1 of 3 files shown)


shared/bundle_analysis/report.py
```diff
@@ -239,12 +239,14 @@ def ingest(self, path: str, compare_sha: Optional[str] = None) -> Tuple[int, str
 
             # Save custom base commit SHA for doing comparisons if available
             if compare_sha:
```
JerrySentry, line 241: This needs to be an upsert, this is one of the only ways of doing it for SQLAlchemy/SQLite

shared/bundle_analysis/comparison.py
```diff
@@ -1,15 +1,21 @@
...
 from shared.bundle_analysis.report import (
     AssetReport,
     BundleAnalysisReport,
     BundleReport,
 )
 from shared.bundle_analysis.storage import BundleAnalysisReportLoader
+from shared.django_apps.core.models import Repository
+from shared.django_apps.reports.models import CommitReport
+
+log = logging.getLogger("__name__")
```
giovanni-guidini, line 18: I don't think you should put the quotes here. Just leave it `__name__` (for the logger to have the name of the file). Right now the logger would be called literally `"__name__"`, I think

giovanni-guidini (approved):
LGTM, left a comment tho.

giovanni-guidini (approved):

PR #333, opened 2024-08-21
chore: adding more causes for InvalidInstallationError

giovanni-guidini:
We are still being hit by many invalid installation errors.
The `permission_error` real cause is still not unclear. At first the suspicion
was suspended apps, but that is not the whole story (from empirical data).

I suspected rate limit errors, and we discovered an unconfigured cache at some point.
That has been remedied but the errors are still going strong.

These changes attempt to further breakdown the responses we get to help with debugging.
Mostly it will verify if the issue is indeed with rate limit or not.

### DIFF
--- a/shared/github/__init__.py
+++ b/shared/github/__init__.py
@@ -39,13 +39,46 @@ def get_pem(*, pem_name: Optional[str] = None, pem_path: Optional[str] = None) -
     raise Exception("No PEM provided to get installation token")
 
 
-InstallationErrorCause = Literal["installation_not_found"] | Literal["permission_error"]
+InstallationErrorCause = (
+    Literal["installation_not_found"]
+    | Literal["permission_error"]
+    | Literal["requires_authentication"]
+    | Literal["installation_suspended"]
+    | Literal["validation_failed_or_spammed"]
+    | Literal["rate_limit"]
+)
 
 
 class InvalidInstallationError(Exception):
     def __init__(self, error_cause: InstallationErrorCause, *args: object) -> None:
         super().__init__(error_cause, *args)
         self.error_cause = error_cause
+
+
+def decide_installation_error_cause(
+    response: requests.Response,
+) -> InstallationErrorCause:
+    # https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-an-installation-access-token-for-an-app
+    is_suspended = (
+        response.json().get("message") == "This installation has been suspended"
+    )
+    is_rate_limit = (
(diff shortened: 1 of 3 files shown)


shared/github/__init__.py
```diff
@@ -48,6 +54,32 @@ def __init__(self, error_cause: InstallationErrorCause, *args: object) -> None:
...
+def decide_installation_error_cause(
+    response: requests.Response,
+) -> InstallationErrorCause:
+    # int(res.headers.get("X-RateLimit-Remaining", -1)) == 0
+    is_primary_rate_limit = int(response.headers.get("X-RateLimit-Remaining", -1)) == 0
+    is_secondary_rate_limit = response.headers.get("Retry-After") is not None
+    match response.status_code:
+        case 404:
+            return "installation_not_found"
+        case 429:
```
Swatinem, line 66: this is unreachable, as you call the function within a `status_code in (404, 403)` block

shared/github/__init__.py
```diff
@@ -48,6 +54,32 @@ def __init__(self, error_cause: InstallationErrorCause, *args: object) -> None:
         self.error_cause = error_cause
 
 
+def decide_installation_error_cause(
+    response: requests.Response,
+) -> InstallationErrorCause:
+    # int(res.headers.get("X-RateLimit-Remaining", -1)) == 0
+    is_primary_rate_limit = int(response.headers.get("X-RateLimit-Remaining", -1)) == 0
+    is_secondary_rate_limit = response.headers.get("Retry-After") is not None
```
Swatinem, line 62: I’m a bit confused by this, are those two not always equal?
As in: `Retry-After` is only set when `X-RateLimit-Remaining == 0`.

Also not quite sure how python `requests` handles header normalization in terms of lowercasing? I would assume it just does the right thing and internally lowercases everything?

giovanni-guidini: About the "lowecasing everything", I got those bits from [here](https://github.com/codecov/shared/blob/gio/debug-installation-error/shared/torngit/github.py#L865)... and that seems to be working... I should test those things just to be sure though.

I don't see why it should lowercase headers that start with `X-*` tho?

shared/github/__init__.py
giovanni-guidini: Thanks for pointing it out

shared/github/__init__.py
giovanni-guidini: I didn't find evidence that `requests` lowercases the headers (testing locally, I mean)

Swatinem (approved):

Swatinem (approved):

PR #334, opened 2024-08-23
chore: mark suspended apps so

giovanni-guidini:
In commit 6637dcc392d7adde051948cb9053201fd0e9cbda we added a further breakdown
of why `InvalidInstallationError` was raised when trying to get the access_token.
In particular the body of the response tells us if the installation is suspended,
which is quite convenient.

Scrolling through the events in Sentry for that issue - [example](https://codecov.sentry.io/issues/5619600100/events/d9d0c0f6e66a43dba92e4d0021b9f0fa/?project=4165036&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=previous-event&statsPeriod=7d&stream_index=10&utc=true) - there's quite a few "installation_suspended" entries.

We do have a mechanism to mark installations as suspended, so we don't try to get tokens for those, but it relies on webhooks, and these installations were probably suspended before that was put in place.

So these changes mark installations that are suspended as suspended after we get the error.

To unsuspend an installation is a manual task done by the user in the GitHub UI. It sends us a webhook and we mark the app as unsuspended.

### DIFF
--- a/shared/bots/github_apps.py
+++ b/shared/bots/github_apps.py
@@ -12,7 +12,11 @@ from shared.django_apps.codecov_auth.models import (
     Service,
 )
 from shared.django_apps.core.models import Repository
-from shared.github import get_github_integration_token, is_installation_rate_limited
+from shared.github import (
+    InvalidInstallationError,
+    get_github_integration_token,
+    is_installation_rate_limited,
+)
 from shared.helpers.redis import get_redis_connection
 from shared.orms.owner_helper import DjangoSQLAlchemyOwnerWrapper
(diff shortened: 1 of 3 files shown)


giovanni-guidini:
In terms of future work we might want to start deleting the installations that are not found

giovanni-guidini:
> lgtm! i would try to document the fact that to resolve this it will require manual work from the user in the Github UI

We do have a ticket to surface it in the UI that mentions it too... Or you think it should be documented directly in the code? (Not counting the commit message that does say that)

PR #335, opened 2024-08-23
feat: Add .mjs and .cjs to AssetType.JAVASCRIPT

suejung-sentry:
Add additional file extensions to supported asset types in Bundle Analysis AssetReport. May cause large jumps to stats as subsequent reports count sizes of files not previously captured.

Paired with: 
* https://github.com/codecov/codecov-api/pull/769
* https://github.com/codecov/worker/pull/644

Issue: https://github.com/codecov/engineering-team/issues/2211


shared/bundle_analysis/parsers/v2.py
```diff
@@ -163,7 +163,7 @@ def parse(self, path: str) -> Tuple[int, str]:
     def _asset_type(self, name: str) -> AssetType:
         extension = get_extension(name)
 
-        if extension in ["js"]:
+        if extension in ["js", "mjs", "cjs"]:
```
JerrySentry, line 166: btw there is also the v1 parser, does this change mean only v2 of the plugin schema can support mjs and cjs?

JerrySentry (approved):

PR #337, opened 2024-08-27
334 v2 adjust installation key

adrian-codecov:
Seeing a couple of issues in Sentry about key error, https://codecov.sentry.io/issues/?groupStatsPeriod=auto&project=4165036&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D+KeyError&referrer=issue-list&statsPeriod=3h&stream_index=6. I tracked it down to this line, https://github.com/codecov/shared/blob/114034600da5b7a3f974a5201b5c51917617bc50/shared/bots/github_apps.py#L323-L324, being called, where the GithubInstallationInfo only returns the installation_id, so I'm using that key instead of the id. I verified this key being unique from GH's side, so we can safely filter by that key. The only drawback is that the installation_id isn't indexed, so the filtering logic might take a bit more time (but we can add an index in a followup PR)

### Legal Boilerplate

### DIFF
--- a/shared/bots/github_apps.py
+++ b/shared/bots/github_apps.py
@@ -143,7 +143,7 @@ def get_github_app_token(
         # The token is not owned by an Owner object, so 2nd arg is None
         return installation_token, None
     except InvalidInstallationError as err:
-        if err.error_cause == "installation_suspended":
+        if err.error_cause == "installation_suspended" and "id" in installation_info:
             # Mark the installation as suspended so we don't keep trying to get the token for it
             GithubAppInstallation.objects.filter(id=installation_info["id"]).update(
                 is_suspended=True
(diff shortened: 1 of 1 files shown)


giovanni-guidini (commented):
Nice job tracking down this bug. Thanks for that.

If you look closer though, that line is from the deprecated flow, during a time that there was no `GithubAppInstallation` object.
So it doesn't return an `id` because there isn't one. There's also no object to mark as suspended.

So the actual fix would be something like

```python
if err.error_cause == "installation_suspended" and "id" in installation_info:
    # Mark the installation as suspended so we don't keep trying to get the token for it
    GithubAppInstallation.objects.filter(
                installation_id=installation_info["id"]
            ).update(is_suspended=True)
```

giovanni-guidini (approved):

PR #346, opened 2024-09-04
bundle analysis: introduce asset ordering in query

JerrySentry:
- Has an option to get list of assets ordered by name or size columns.
- Add `id` attribute to asset report, this will be useful for implementing pagination on the api side.

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/report.py
+++ b/shared/bundle_analysis/report.py
@@ -5,7 +5,7 @@ import sqlite3
 import tempfile
 from typing import Any, Dict, Iterator, List, Optional, Set, Tuple
 
-from sqlalchemy import text
+from sqlalchemy import asc, desc, text
 from sqlalchemy.exc import OperationalError
 from sqlalchemy.orm import Session as DbSession
 from sqlalchemy.orm.query import Query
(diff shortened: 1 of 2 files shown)


shared/bundle_analysis/report.py
```diff
@@ -129,8 +133,11 @@ def asset_reports(
         asset_types: Optional[List[AssetType]] = None,
         chunk_entry: Optional[bool] = None,
         chunk_initial: Optional[bool] = None,
+        ordering_column: Optional[str] = "size",
```
JerrySentry, line 136: Leaving it to callers that use `BundleReport` to validate existence of columns

shared/bundle_analysis/report.py
```diff
@@ -142,7 +149,7 @@ def asset_reports(
                 asset_types,
                 chunk_entry,
                 chunk_initial,
-            )
+            ).order_by(ordering(getattr(Asset, ordering_column)))
```
giovanni-guidini, line 152: Can you explain the purpose of this `getattr` there?... You are grabbing the value from the column listed in `ordering_column`. Is that the same as a reference to the column?

giovanni-guidini (approved):
LGTM but I didn't understand some implementation details.

shared/bundle_analysis/report.py
JerrySentry: It's doing something like `order_by(getattr(Asset, "size"))`, which is really just doing a `order_by(Asset.size)`.

PR #347, opened 2024-09-05
Avoid metadata requests talking to GCS

Swatinem:
The GCS Storage would previously fetch the bucket metadata using `get_bucket`, where a call to `bucket` is sufficient to create the python class without doing a request.

---

fixes https://github.com/codecov/engineering-team/issues/2276


joseph-sentry (dismissed):

Swatinem:
Interesting, I now noticed the metadata updates in a trace for the first time:

image

Swatinem:
I updated this with some of the learnings from https://github.com/codecov/shared/pull/351.

This should now be a "minimally invasive" version.

It avoids the needless bucket metadata calls, but otherwise keeps the file metadata call as-is because of the gzip oddity.

I also hand-edited the `cassette` files to remove the bucket metadata calls from them, and to remove some obsolete ones, but the tests would function even without that update.

michelletran-codecov (approved):

matt-codecov (approved):

PR #349, opened 2024-09-05
bundle analysis: add gzip size total for report

JerrySentry:
Add a function for BundleReport to fetch gzip size total

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/report.py
+++ b/shared/bundle_analysis/report.py
@@ -3,7 +3,7 @@ import logging
 import os
 import sqlite3
 import tempfile
-from typing import Any, Dict, Iterator, List, Optional, Set, Tuple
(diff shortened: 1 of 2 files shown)


shared/bundle_analysis/report.py
```diff
@@ -128,18 +128,16 @@ def _asset_filter(
             query = query.filter(Asset.asset_type.in_(asset_types))
         return query
```
JerrySentry, line 130: Basically the code change went from:
```
def total_size:
    return query(column=Asset.size)
```
to
```
def _get_size(column):
    return query(column=column)

def total_size:
    return _get_size(Asset.size)

def total_gzip_size:
    return _get_size(Asset.gzip_size)
```

PR #353, opened 2024-09-09
feat: add percentage change for bundles

giovanni-guidini:
The design for https://github.com/codecov/engineering-team/issues/1491 includes warnings in the bundle table for individual bundles.

In order to do that we need a percentage change on the bundle level. The calculation of percentage follows the same idea as the `Comparison`: the percentage is taken using the base bundle size as the 100%

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/comparison.py
+++ b/shared/bundle_analysis/comparison.py
@@ -44,6 +44,7 @@ class BundleChange:
     bundle_name: str
     change_type: ChangeType
     size_delta: int
+    percentage_delta: float  # [-100, 100] range
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


shared/bundle_analysis/comparison.py
```diff
@@ -44,6 +44,7 @@ class ChangeType(Enum):
     bundle_name: str
     change_type: ChangeType
     size_delta: int
+    percentage_delta: float  # [-100, 100] range
```
Swatinem, line 47: I don’t see any code that is explicitly clamping this to the said range. does that happen someplace thats not visible in this diff?

Swatinem (dismissed):

shared/bundle_analysis/comparison.py
giovanni-guidini: Hum good point, not really. I'll remove the comment, as it would not be possible to clamp it and still have accurate numbers I guess

Swatinem (approved):

PR #357, opened 2024-09-13
Advertise `shared` as being typed, and add vscode settings

Swatinem:
The main addition here is the `py.typed` file, which advertises to `mypy` that this package indeed has type annotations, so downstream users like `worker` are properly using those.

As another change, this adds some vscode settings with recommended extensions, and sets up ruff to auto-fix/format on-save, which is copied from https://github.com/codecov/worker/pull/706.


Swatinem:
It seems that I am also running into https://github.com/python/mypy/issues/13392, as I can’t `pip install --editable ../shared` from worker unfortunately :-(

I am once again hitting the limit of my knowledge of the python ecosystem here.

giovanni-guidini (approved):

PR #360, opened 2024-09-17
feat: create manual partitions for backfilling daily test rollups

joseph-sentry:


michelletran-codecov (approved):
This look good because it looks like you're just adding the partition tables and not doing any data migration. What is the plan for the backfill?

joseph-sentry:
> What is the plan for the backfill?
One off celery task that will go through each repo that has test analytics enabled and backfill the rollups (this will overwrite the rollups if they already exist)

PR #364, opened 2024-09-19
fix: use update_or_create instead of get_or_create depending on the webhook response

ajay-sentry:
This PR aims to fix these related sentry issues:

https://codecov.sentry.io/issues/5334155312
https://codecov.sentry.io/issues/5409774549
https://codecov.sentry.io/issues/5215132965
https://codecov.sentry.io/issues/5820085486
https://codecov.sentry.io/issues/5441235127

That seem to all be related to a unique_constraint error when trying to create this object when a similar one already exists. Thanks @adrian-codecov for talking me through this one

Unique constraint comes from here: https://github.com/codecov/shared/blob/b42ecbe84e6cf1e0d8cfdab6dc5beb12600e06b3/shared/django_apps/core/models.py#L157-L163

Related django documentation: https://docs.djangoproject.com/en/5.1/ref/models/querysets/#update-or-create

### Legal Boilerplate

### DIFF
--- a/shared/django_apps/core/managers.py
+++ b/shared/django_apps/core/managers.py
@@ -285,13 +285,26 @@ class RepositoryQuerySet(QuerySet):
     def get_or_create_from_git_repo(self, git_repo, owner):
         from shared.django_apps.codecov_auth.models import Owner
 
-        repo, created = self.update_or_create(
-            author=owner,
-            service_id=git_repo.get("service_id") or git_repo.get("id"),
-            private=git_repo["private"],
-            branch=git_repo.get("branch") or git_repo.get("default_branch") or "main",
-            name=git_repo["name"],
-        )
+        service_id = git_repo.get("service_id") or git_repo.get("id")
+        name = git_repo["name"]
+
+        defaults = {
+            "private": git_repo["private"],
+            "branch": git_repo.get("branch")
+            or git_repo.get("default_branch")
+            or "main",
+            "name": name,
+            "service_id": service_id,
+        }
+
+        try:
+            repo, created = self.update_or_create(
+                author=owner, service_id=service_id, defaults=defaults
+            )
+        except IntegrityError:
+            repo, created = self.update_or_create(
+                author=owner, name=name, defaults=defaults
+            )
 
         # If this is a fork, create the forked repo and save it to the new repo.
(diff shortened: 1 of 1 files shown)


shared/django_apps/core/managers.py
```diff
@@ -284,7 +285,7 @@ def with_oldest_commit_at(self):
     def get_or_create_from_git_repo(self, git_repo, owner):
         from shared.django_apps.codecov_auth.models import Owner
 
-        repo, created = self.get_or_create(
```
JerrySentry, line 287: I'm not sure how this fixes it, so the unique constraints are the triplet of `service_id`, `name`, and `author `right? And I'm guessing the initial error happens when `get_or_create` is called with the same above triplet but different `branch` or `private` values. It doesn't recognize these 5 args being in the DB so it tries to create a new one, but failing from the unique constraints. This is what's currently happening right?

Now with `update_or_create`, similarly it doesn't find these 5 args in the DB so it won't update it would do create leading to the same error no?

JerrySentry: I'm thinking the way to do it is one of these 2:
1. Ignore changes on subsequent changes to `branch` and `private` columns
```
repo, created = self.get_or_create(
    author=owner,
    service_id=git_repo.get("service_id") or git_repo.get("id"),
    name=git_repo["name"],
    defaults={
        "branch": git_repo.get("branch") or git_repo.get("default_branch") or "main",
        "private": git_repo["private"],
    }
)
```
1. Updates changes on subsequent changes to `branch` and `private` columns
```
repo, created = self.update_or_create(
    author=owner,
    service_id=git_repo.get("service_id") or git_repo.get("id"),
    name=git_repo["name"],
    defaults={
        "branch": git_repo.get("branch") or git_repo.get("default_branch") or "main",
        "private": git_repo["private"],
    }
)
```

JerrySentry: Actually that's not gonna work because the unique constraints are 2 pairs and `get_or_update` is 1 triplet. So if you had in the DB `(author=codecov, service_id=1, name=gazebo)`
and call with params:
`(author=codecov, service_id=2, name=gazebo)`
a unique constraint error would occur as well because (author, name) constraint fail.

JerrySentry (dismissed):

JerrySentry (approved):

PR #365, opened 2024-09-20
Make dailytestrollup branch nullable

michelletran-codecov:
Admittedly, I'm not sure if this is the right solution, but we've been seeing a bunch of errors related to this field not being nullable (see [error](https://codecov.sentry.io/share/issue/478a3e60158547b69afa0b9959822237/) in Sentry). This is because it's [taking the branch from the commit](https://github.com/codecov/worker/blob/a7cb347f36b0fc45cfe29ac1390a7b58020f4dcd/tasks/test_results_processor.py#L288), which _can_ be nullable.

This would also fix a bunch of errors seen in https://github.com/codecov/worker/pull/729

### Legal Boilerplate


shared/django_apps/reports/models.py
```diff
@@ -404,7 +404,7 @@ class PartitioningMeta:
     )
     date = models.DateField()
     repoid = models.IntegerField()
-    branch = models.TextField()
+    branch = models.TextField(null=True)
```
adrian-codecov, line 407: @joseph-sentry I'll summon thee to see if it'd affect your work

joseph-sentry: ah this is fine i think, it's probably crashing the test results processor at the moment, I find it weird that we have commits with no branch but i guess that's a separate issue

joseph-sentry (approved):

joseph-sentry (approved):

PR #373, opened 2024-09-27
add upload_token_required_for_public_repos, backfill and make non-nullable

nora-codecov:
https://github.com/codecov/engineering-team/issues/2297


Swatinem (approved):
lgtm.
So this will create the column as `NULL`, the backfill sets that to `true`, and then later on the `NULL` is removed, and the default `false` remains?

RulaKhaled (approved):

PR #377, opened 2024-10-02
Add new field email_opt_in to user model

RulaKhaled:
This is needed for marketing reasons. 

### Legal Boilerplate


rohitvinnakota-codecov (dismissed):

shared/django_apps/codecov_auth/models.py
```diff
@@ -87,6 +87,7 @@ class CustomerIntent(models.TextChoices):
     terms_agreement = models.BooleanField(null=True, default=False, blank=True)
     terms_agreement_at = DateTimeWithoutTZField(null=True, blank=True)
     customer_intent = models.TextField(choices=CustomerIntent.choices, null=True)
+    email_opt_in = models.BooleanField(null=True, default=False, blank=True)
```
nora-codecov, line 90: Hi Rola! You don't need to set `blank` here.

nora-codecov: also, do you want to make this nullable - does `None` have a business logic meaning in this field?

RulaKhaled: i guess not, will remove

nora-codecov (approved):
One last question - should this be a risky migration? I'm not sure how big this table is.

make sure you don't make the same mistake I did, where I put api and worker's representation of the table out of sync 🙃 

arpad fixed it here https://github.com/codecov/worker/pull/757
(the default value lives in django, so worker's sqlalchemy and the database's psql don't know what the default should be, so they try to use null and get an error because the field is not nullable)

RulaKhaled:
> One last question - should this be a risky migration? I'm not sure how big this table is.

i think we don't usually mark user related fields as risky, thank you for the heads up!

PR #381, opened 2024-10-09
chore: use graphQL to get PR info (github)

giovanni-guidini:
Changes the endpoint to get a single PR info to use graphQL api.
The goal is to slightly improve the baseline time of the request and
reduce the difference a little from the fastest request to p90.

For more details on the exploration see https://github.com/codecov/engineering-team/issues/2627

Other changes are to improve some typings a little bit.
It also introduces the `ProviderPull` response type.

Unfortunately it was a bit overkill exploration for what are really small gains 😅

### Legal Boilerplate


shared/torngit/response_types.py
```diff
@@ -0,0 +1,24 @@
...
+
+class ProviderAuthor(TypedDict):
+    id: str | None
+    username: str | None
+
+
+class ProviderCommit(TypedDict):
+    branch: str
+    commitid: str
+    slug: NotRequired[str]  # Only GitHub includes slug
```
Swatinem, line 12: interesting, first time I learn about `NotRequired`

Swatinem (approved):
lgtm

when reading the description, I thought this would be more about merging more than one request into a graphql query, like I suggested in https://github.com/codecov/engineering-team/issues/2357 as well, though it looks like this only replaces a single request.

giovanni-guidini:
> lgtm
> 
> when reading the description, I thought this would be more about merging more than one request into a graphql query, like I suggested in [codecov/engineering-team#2357](https://github.com/codecov/engineering-team/issues/2357) as well, though it looks like this only replaces a single request.

I think this brings us closer to having the commits return in a single request (because I think the pull request info in the GraphQL can return that too), but for this particular PR I wanted to have as minimal of a change as possible :E drop in replacement for that 1 request

PR #382, opened 2024-10-09
Adding actions to comments

adrian-codecov:
### Legal Boilerplate


trent-codecov (approved):

shared/torngit/github.py
```diff
@@ -609,6 +610,33 @@ def count_and_get_url_template(cls, url_name):
...
+                    "github", "comment_action_bot_name", default="sentry-ai"
+                )
+                body["actions"] = [
+                    {
+                        "name": "Open Sentry Agent",
+                        "type": "copilot-chat",
+                        "prompt": f"@{bot_name} generate tests for PR",
+                    }
+                ]
+        except Exception:
```
nora-codecov, line 635: for a try-catch that's this broad, it would be a good idea to log the exceptions with a distinct name, so you can monitor and look up whether any exceptions are happening here

PR #385, opened 2024-10-11
revert 6c0c2d68a0ab914242a8a03bedfcedf7a687588f

giovanni-guidini:
This reverts commit 6c0c2d68a0ab914242a8a03bedfcedf7a687588f.

I'm doing this because this PR got deployed, but the graphQL api somethimes doesn't return the head repo (for example if it was deleted).

We use the head-slug and the base-slug to know if the PR is from a fork. Because the graphQL api includes a `isCrossRepository` bool everytime we can still know that, so there is a future where this work can be used.

However this will take some more time and I don't want to clog `shared` in the meantime. So it's best to revert for now, to remove these changes from  `main`.

### Legal Boilerplate


Swatinem (approved):
there were some nice improvements to typing overall, it might be nice to do those changes separately

PR #394, opened 2024-10-17
Update YAML validator to drop reserved key for internal use only

JerrySentry:
This change disallows the `to_string` key to be present during the validation of codecov YAML, if whenever the key is in the YAML string then it will fail with validation error. We need this reserved key in order to keep a copy of the owner YAML's full string content allowing to display original formatting and comments.
### Legal Boilerplate

### DIFF
--- a/shared/yaml/validation.py
+++ b/shared/yaml/validation.py
@@ -61,6 +61,11 @@ def pre_process_yaml(inputted_yaml_dict):
             val = inputted_yaml_dict["codecov"]["notify"].pop("require_ci_to_pass")
             inputted_yaml_dict["codecov"]["require_ci_to_pass"] = val
 
+    # Remove the reserved `to_string` key word in the base level of the json.
+    # This key is used to store the full YAML string preserving the original comments.
+    if "to_string" in inputted_yaml_dict:
+        inputted_yaml_dict.pop("to_string")
+
 
 def _calculate_error_location_and_message_from_error_dict(error_dict):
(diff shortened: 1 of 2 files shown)


shared/yaml/validation.py
```diff
@@ -61,6 +61,11 @@ def pre_process_yaml(inputted_yaml_dict):
             val = inputted_yaml_dict["codecov"]["notify"].pop("require_ci_to_pass")
             inputted_yaml_dict["codecov"]["require_ci_to_pass"] = val
 
+    # Remove the reserved `to_string` key word in the base level of the json.
+    # This key is used to store the full YAML string preserving the original comments.
+    if "to_string" in inputted_yaml_dict:
+        inputted_yaml_dict.pop("to_string")
```
suejung-sentry, line 67: Thoughts on adding a helper `remove_reserved_keys` that lets us pop any reserved data in the future?

Something like below?

```

yaml_top_level_reserved_keys: List[str] = ["to_string"]

def remove_reserved_keys(input_dict: Dict[str, any], keys_to_remove: List[str]) -> Dict[str, any]:
    copied_dict = copy.deepcopy(input_dict)
    
    for key in keys_to_remove:
        if key in copied_dict:
            copied_dict.pop(key)
    
    return copied_dict
```

suejung-sentry (approved):
Sweet! LGTM with an optional comment

suejung-sentry (approved):
Nice! Approved it!

PR #397, opened 2024-10-17
Add hide codecov tokens field

rohitvinnakota-codecov:
Adds a YAML field to hide codecov token to non-admin users. A follow up PR will be added in API that reads the config before sending token values via GQL. 

### Legal Boilerplate

### DIFF
--- a/shared/validation/install.py
+++ b/shared/validation/install.py
@@ -130,12 +130,7 @@ config_schema = {
             "api_cors_allowed_origins": {"type": "string"},
             "codecov_dashboard_url": {"type": "string"},
             "enterprise_license": {"type": "string"},
-            "hide_all_codecov_tokens": {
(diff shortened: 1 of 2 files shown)


shared/validation/install.py
```diff
@@ -130,6 +130,12 @@ def check_task_config_key(field, value, error):
             "api_cors_allowed_origins": {"type": "string"},
             "codecov_dashboard_url": {"type": "string"},
             "enterprise_license": {"type": "string"},
+            "hide_all_codecov_tokens": {
```
spalmurray-codecov, line 133: why not just a boolean type?

rohitvinnakota-codecov: yeah can do that

spalmurray-codecov (approved):

PR #405, opened 2024-10-21
feat: use zstd instead of gzip for compression

joseph-sentry:
- use zstandard compression in write_file
- if the content encoding is set to zstd, manually decompress the downloaded bytes using ztd
- remove logic from renaming content_type in read_file
- setup VCR in test_gcp to match on the body


tests/unit/storage/test_gcp.py
```diff
@@ -1,16 +1,73 @@
...
+            body = [line for line in body if not line.startswith(b"--===============")]
+            body = b"\n".join(body)
+            request.body = body
+        except Exception as e:
+            raise e
+
+        # there's also random data in gzip compressed data so we decompress it
+        if (start := request.body.find(b"\x1f\x8b\x08\x00")) > 0:
+            compressed_body = request.body[start:]
+            request.body = request.body[:start] + gzip.decompress(compressed_body)
```
Swatinem, line 36: there shouldn’t be any "random data" here. I would rather not mess with these fixtures, as I have little confidence in these to begin with :-D

tests/unit/storage/test_gcp.py
```diff
@@ -100,13 +152,15 @@ def test_manually_then_read_then_write_then_read_file(self, codecov_vcr):
...
         storage = GCPStorageService(gcp_config)
         path = "test_write_then_read_file/result"
-        data = gzip.compress("lorem ipsum dolor test_write_then_read_file á".encode())
-        bucket_name = "testingarchive02"
+        data = zstandard.compress(
+            "lorem ipsum dolor test_write_then_read_file á".encode()
+        )
+        bucket_name = "testingarchive1"
         writing_result = storage.write_file(
             bucket_name, path, data, is_already_gzipped=True
```
Swatinem, line 165: I would rather keep the meaning of the `is_already_gzipped` flag, and assume it is indeed `gzip`, as some API users might use it that way.

tests/unit/storage/test_gcp.py
```diff
@@ -119,32 +173,41 @@ def test_write_then_read_file_gzipped(self, codecov_vcr):
...
+        bucket_name = "testingarchive1"
         content_to_upload = "content to write\nThis is crazy\nWhy does this work"
         bucket = storage.storage_client.get_bucket(bucket_name)
         blob = google_storage.Blob(path, bucket)
-        with io.BytesIO() as f:
-            with gzip.GzipFile(fileobj=f, mode="wb", compresslevel=9) as fgz:
-                fgz.write(content_to_upload.encode())
-            blob.content_encoding = "gzip"
-            blob.upload_from_file(
-                f, size=f.tell(), rewind=True, content_type="application/x-gzip"
```
Swatinem, line 138: I don’t think the "file" part of this test is important, but the `application/x-gzip` part I believe is. When I was messing with this code, I found out that the crc32 check was failing depending on the content_type, and I believe that is the reason there was some workaround for it specifically in the code.
We should make sure things are still working as expected there.

Swatinem (commented):
As we have this abstraction in place for all the different storage backends, we should add support for this to the other variants as well.
Reason is that API is using the minio backend to access GCS. I have no idea how that even works, but thats the code that runs in production, and we have to make sure things work out correctly using that.

Other than that, I left some individual comments mostly in the tests about:
- The `is_already_gzipped` flag should assume `gzip`
- We should verify that the `application/x-gzip` handling works properly
- This applies zstd decompression only in the "return as `bytes`" codepath, but not the "save to file".

tests/unit/storage/test_gcp.py
joseph-sentry: the random data is the timestamp in in the gzip compression, afaik there's no way to avoid having it there, but basically it causes the VCR matching for requests to fail because the body of the request _is_ different

i think this is necessary if we want to ensure that the body of the requests is correct, which is how we ensure that the compression is occuring properly

tests/unit/storage/test_gcp.py
joseph-sentry: having a separate `is_already_zstded` feels bad to me because then would we need some validation to make sure that both `is_already_gzipped` and `is_already_zstded` are both not set to True, and to me it would just make the interface more complicated

i did want to make this: `is_already_compressed` but that would require a bunch of changes wherever this is being used because it's a kwarg

tests/unit/storage/test_gcp.py
joseph-sentry: this makes sense and i will take a look

Swatinem:
We have gone a few iterations on the archive service, and there is a bunch of PRs open:
- this one
- #351 
- #415

Now the circumstances have changed a little in the sense that it looks like we might want to consolidate the storage layer as only using `minio` as the storage adapter, as that already runs against the GCS server in `api` and staging `worker`.
So we can just remove all the other storage adapters.

IMO, we could:
- take #415 as that is some general cleanup and reduces the surface area (I would like to land this together with https://github.com/codecov/codecov-api/pull/960, as that changes the user of said APIs)
- remove all the other storage adapters, moving exclusively to `minio`
- take the remaining parts of #351, which would remove all the cumbersome `vcr` tests and replace those with running against a live minio server in docker
- adapt this PR to add a better story around compression, and move the default to `zstd`
- then as a followup, we can also consolidate the `ArchiveService` layer across `worker`, `api` and `shared` (the latter two should be handled by https://github.com/codecov/codecov-api/pull/960)

shared/storage/minio.py
```diff
@@ -143,50 +162,64 @@ def create_root_storage(self, bucket_name="archive", region="us-east-1"):
...
+        data: BinaryIO,
+        reduced_redundancy: bool = False,
         *,
-        is_already_gzipped: bool = False,
+        is_already_gzipped: bool = False,  # deprecated
+        is_compressed: bool = False,
+        compression_type: str = "zstd",
     ):
+        if is_already_gzipped:
+            log.warning(
```
Swatinem, line 175: this would spam warnings for *all* of the calls setting this flag.
As it should be fairly trivial to migrate code over to the new flags, this should probably not log any warnings or sentry messages, but just perform the fallback.

Ideally the migration path would be 4 PRs:
- this one
- migrate `is_already_gzipped` users within `worker`
- migrate `is_already_gzipped` users within `api`
- remove the flag and backwards compatibility code

shared/storage/minio.py
```diff
@@ -143,50 +162,64 @@ def create_root_storage(self, bucket_name="archive", region="us-east-1"):
...
+            with sentry_sdk.new_scope() as scope:
+                scope.set_extra("bucket_name", bucket_name)
+                scope.set_extra("path", path)
+                sentry_sdk.capture_message("is_already_gzipped passed with True")
+            is_compressed = True
+            compression_type = "gzip"
+
         if isinstance(data, str):
-            data = data.encode()
+            log.warning(
```
Swatinem, line 186: same here. it should be simple enough (famous last words and all, lol) to audit all users of this API and migrate them in a 3-step process (4 as we have 2 dependent repos).

shared/storage/minio.py
```diff
@@ -143,50 +162,64 @@ def create_root_storage(self, bucket_name="archive", region="us-east-1"):
...
+            _, filepath = tempfile.mkstemp()
+            with open(filepath, "wb") as f:
+                while chunk := reader.read(16384):
+                    f.write(chunk)
+            data = open(filepath, "rb")
 
         try:
-            # reset pos for minio reading.
-            out.seek(0)
+            out_size = data.seek(0, os.SEEK_END)
```
Swatinem, line 206: does minio require the size to be known ahead of time?
if not, we wouldn’t need to compress into a temporary file at all, and could just pass the compression stream directly, so it would do proper streaming compression during the upload.

shared/storage/minio.py
```diff
@@ -195,25 +228,65 @@ def write_file(
...
+                reader = data
+            elif content_encoding == "zstd":
+                # we have to manually decompress zstandard compressed data
+                cctx = zstandard.ZstdDecompressor()
+                reader = cctx.stream_reader(data)
+            else:
+                with sentry_sdk.new_scope() as scope:
+                    scope.set_extra("bucket_name", bucket_name)
+                    scope.set_extra("path", path)
+                    raise ValueError("Blob does not have Content-Encoding set")
```
Swatinem, line 279: I don’t think this should be a hard error. As uploading is the responsibility of the client, they can choose not to compress at all, or use another compression method that is transparently handled by `Content-Encoding` and `HTTPResponse`.

shared/storage/minio.py
```diff
@@ -195,25 +228,65 @@ def write_file(
...
-                data = BytesIO()
-                for d in res.stream(CHUNK_SIZE):
-                    data.write(d)
-                data.seek(0)
-                return data.getvalue()
+            headers = {"Accept-Encoding": "gzip, zstd"}
+            if file_obj:
+                _, tmpfilepath = tempfile.mkstemp()
+                to_file_response: GetObjectToFileResponse = (
+                    self.minio_client.fget_object(
```
Swatinem, line 250: I’m not sure we need these two different codepaths at all, given that you stream the response through decompression anyway, and then collect it into the `file_obj` later on anyway. This should avoid having to handle tempfiles completely.

shared/storage/minio.py
```diff
@@ -195,25 +228,65 @@ def write_file(
...
+            else:
+                response: HTTPResponse = self.minio_client.get_object(
+                    bucket_name, path, request_headers=headers
+                )
+                data = response
+                content_encoding = response.headers.get("Content-Encoding", None)
+
+            reader: Readable | None = None
+            if content_encoding == "gzip":
+                # HTTPResponse automatically decodes gzipped data for us
```
Swatinem, line 267: I wonder if an uptodate `urllib3` would also handle `zstd` automatically for us.
We have quite an outdated `urllib3` dependency primarily for reasons of compatiblity with `vcr` snapshots.

I’m pretty sure that `Content-Encoding: zstd` is properly standardized and transparently handled by browsers by now for example.

shared/storage/minio.py
```diff
@@ -195,25 +228,65 @@ def write_file(
...
+            else:
+                with sentry_sdk.new_scope() as scope:
+                    scope.set_extra("bucket_name", bucket_name)
+                    scope.set_extra("path", path)
+                    raise ValueError("Blob does not have Content-Encoding set")
+
+            if file_obj:
+                while chunk := reader.read(16384):
+                    file_obj.write(chunk)
+                return None
```
Swatinem, line 284: I believe a `file_obj.seek(0)` is needed here

Swatinem:
Another thing I remembered is that we should add a test that `Content-Encoding: gzip; Content-Type: application/x-gzip` works correctly.
The gcp backend had a workaround specifically for that. I suspect that whatever http abstraction that was using was not doing the transparent decompression correctly, and also triggered a bogus checksum mismatch error in that case.

shared/storage/minio.py
joseph-sentry: i believe the call to put_object requires the size to be provided

shared/storage/minio.py
```diff
@@ -195,25 +211,46 @@ def write_file(
...
 
     @overload
-    def read_file(self, bucket_name: str, path: str) -> bytes: ...
+    def read_file(
+        self, bucket_name: str, path: str, file_obj: None = None
+    ) -> bytes: ...
 
     @overload
-    def read_file(self, bucket_name: str, path: str, file_obj: BinaryIO) -> None: ...
+    def read_file(self, bucket_name: str, path: str, file_obj: str) -> None: ...
```
Swatinem, line 225:
```suggestion
    def read_file(self, bucket_name: str, path: str, file_obj: BinaryIO) -> None: ...
```

shared/storage/minio.py
```diff
@@ -17,13 +17,29 @@
...
+
+class GetObjectToFileResponse(Protocol):
+    bucket_name: str
+    object_name: str
+    last_modified: datetime.datetime | None
+    etag: str
+    size: int
+    content_type: str | None
+    metadata: dict[str, str]
+    version_id: str | None
```
Swatinem, line 40:
```suggestion
```

This is now unused

shared/storage/minio.py
```diff
@@ -17,13 +17,29 @@
 )
 from minio.deleteobjects import DeleteObject
 from minio.error import MinioException, S3Error
+from urllib3.response import HTTPResponse
 
-from shared.storage.base import CHUNK_SIZE, BaseStorageService
```
Swatinem, line 21: you could keep the `CHUNK_SIZE` constant instead of hardcoding 16k in all the places.

shared/storage/minio.py
```diff
@@ -143,50 +160,49 @@ def create_root_storage(self, bucket_name="archive", region="us-east-1"):
...
-                    shutil.copyfileobj(data, f)
-                out = open(filename, "rb")
-            else:
-                out = data
+        if isinstance(data, str):
+            data = BytesIO(data.encode())
 
-            out_size = os.stat(filename).st_size
+        if not is_compressed:
+            cctx = zstandard.ZstdCompressor()
```
Swatinem, line 180: this would currently break calling this with `compression_type="gzip"`, as it would default to zstd, and then define the wrong header.

IMO we should also support a `compression="none"` flag which does not compress at all, and also does not add any `Content-Encoding` header.

shared/storage/minio.py
```diff
@@ -1,29 +1,42 @@
...
+    def __init__(self, fileobj: IO[bytes]):
+        self.data = fileobj
+
+    def read(self, size: int = -1, /) -> bytes:
+        curr_data = self.data.read(size)
+
+        if not curr_data:
+            return b""
+
+        return gzip.compress(curr_data)
```
Swatinem, line 37: this means you are compressing each chunk you read individually, which is not ideal from a compression point of view, but also means you end up with multiple concatenated gzip streams. I know zstd can handle that, though depending on the specific library you use, you have to opt-into this behavior. I’m not sure about gzip in this case.

I’m surprised you have to implement this yourself though. I would imagine gzip should have a streaming compressor built in?

shared/storage/minio.py
```diff
@@ -143,85 +157,118 @@ def create_root_storage(self, bucket_name="archive", region="us-east-1"):
...
                     f"File {path} does not exist in {bucket_name}"
                 )
             raise e
-        except MinioException:
-            raise
+        if response.headers:
+            content_encoding = response.headers.get("Content-Encoding", None)
+            if content_encoding == "zstd":
+                # we have to manually decompress zstandard compressed data
+                cctx = zstandard.ZstdDecompressor()
```
Swatinem, line 247: Another thing I looked up yesterday:
Since this is builtin with `urllib3 > 2`, you could check `urlllib3.__version__` to make sure we only manually decompress with version 1.
Or maybe `urllib3` has a different flag to check whether it handles zstd internally?

Otherwise we end up in a situation where we update the dependency and then we try to decompress uncompressed bytes and then fail.

shared/storage/minio.py
joseph-sentry: > this means you are compressing each chunk you read individually, which is not ideal from a compression point of view, but also means you end up with multiple concatenated gzip streams. I know zstd can handle that, though depending on the specific library you use, you have to opt-into this behavior. I’m not sure about gzip in this case.

this seems to be fine in the case of gzip: [link](https://www.gnu.org/software/gzip/manual/html_node/Advanced-usage.html)

> I’m surprised you have to implement this yourself though. I would imagine gzip should have a streaming compressor built in?

it doesn't follow the same interface as the python-zstandard streaming interface, and i couldn't come up with a way to code it that isn't awkward.

Swatinem (approved):
nice, this looks great!

feel free to add a test for gzip chunking, and/or add an explicit check for the `urllib3` version.
otherwise its also fine to do those as followups, and lets ship this (carefully to staging first :-)

PR #407, opened 2024-10-23
feat: add inc_counter function to handle errors with metrics

tony-codecov:
Adding an `inc_counter` function to handle errors with increment Prometheus counters.

### Legal Boilerplate


shared/metrics/__init__.py
```diff
@@ -1,5 +1,9 @@
+import logging
+
 from prometheus_client import Counter, Histogram, Summary, start_http_server
 
+log = logging.getLogger(__name__)
```
tony-codecov, line 5: @michelletran-codecov 
The new logger `_name_` would probably be `shared.shared.metrics` so do you think it would be harder to trace back to where the error occured knowing the `_name` of the Counter only?

shared/metrics/__init__.py
```diff
@@ -9,3 +13,10 @@
     "Summary",
     "start_prometheus",
 ]
+
+
+def inc_counter(counter: Counter, labels: dict):
```
michelletran-codecov, line 18:
```suggestion
def inc_counter(counter: Counter, labels: dict) -> None:
```

michelletran-codecov (dismissed):

michelletran-codecov (approved):

PR #408, opened 2024-10-23
feat: Add gauges to metrics and support counters with no labels

tony-codecov:
Add gauges to metrics and support counters with no labels

### Legal Boilerplate

### DIFF
--- a/shared/metrics/__init__.py
+++ b/shared/metrics/__init__.py
@@ -16,7 +16,7 @@ __all__ = [
 ]
 
 
-def inc_counter(counter: Counter, labels: dict = None) -> None:
+def inc_counter(counter: Counter, labels: dict | None = None) -> None:
     try:
         if labels:
             counter.labels(**labels).inc()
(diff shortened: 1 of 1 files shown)


shared/metrics/__init__.py
```diff
@@ -9,14 +9,28 @@
...
     "Counter",
+    "Gauge",
     "Histogram",
     "Summary",
     "start_prometheus",
 ]
 
 
-def inc_counter(counter: Counter, labels: dict) -> None:
+def inc_counter(counter: Counter, labels: dict = None) -> None:
```
Swatinem, line 19:
```suggestion
def inc_counter(counter: Counter, labels: dict | None = None) -> None:
```

Same below. The default value has to match the correct type.

tony-codecov: good catch, thanks!

Swatinem (approved):
code lgtm.

though I’m not convinced having these `inc_counter` / `set_gauge` methods provides any value at all over just using the prometheus API directly.

PR #411, opened 2024-10-29
add FieldTracker to Owner - no migration required!

nora-codecov:
We have this on `Repository` model so we can do [this](https://github.com/codecov/codecov-api/blob/f48452449f88a23715104f3bc17027d7c094563b/core/signals.py#L27) in `api`.
I'd like to add the same thing for `Owner` to use in the same way in `api`.

ChatGPT says it's performant 👼 

https://github.com/codecov/engineering-team/issues/2299

### DIFF
--- a/shared/django_apps/codecov_auth/models.py
+++ b/shared/django_apps/codecov_auth/models.py
@@ -16,6 +16,7 @@ from django.db.models.manager import BaseManager
 from django.forms import ValidationError
 from django.utils import timezone
 from django_prometheus.models import ExportModelOperationsMixin
+from model_utils import FieldTracker
 
 from shared.config import get_config
 from shared.django_apps.codecov.models import BaseCodecovModel, BaseModel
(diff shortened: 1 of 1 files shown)


shared/django_apps/codecov_auth/models.py
```diff
@@ -381,7 +382,9 @@ class Meta:
     )
 
     objects = OwnerManager()
-
+    tracker = FieldTracker(
+        fields=["username", "service", "upload_token_required_for_public_repos"]
```
adrian-codecov, line 386: This is fine, although if you don't specify the fields, then you can track any field you want, so you wouldn't have to make a future PR for this. Pros and cons, this makes it more specific too. I'll leave this up to you!

nora-codecov: yeah I hear that - at the moment I'd like to keep it as slim as possible, so I'll keep it with these specific fields.

adrian-codecov (approved):

PR #414, opened 2024-11-05
fix: add unique constraint to test flag bridge

joseph-sentry:


Swatinem (approved):
lgtm, though have you checked already if the migration will succeed at all? so is the existing data already unique and won’t cause conflicts right away?

joseph-sentry:
> lgtm, though have you checked already if the migration will succeed at all? so is the existing data already unique and won’t cause conflicts right away?

indeed i checked that there would be no conflicts

marcusramberg:
In our self hosted instance, we had 16 duplicates in 

``` sql
SELECT id
    FROM reports_test_results_flag_bridge
    WHERE ID NOT IN
    (
        SELECT MAX(ID)
        FROM reports_test_results_flag_bridge
        GROUP BY test_id, 
                 flag_id
    );
```

Which made the migration fail and report upload subsequently fail on the ON CONFLICT clause. Had to delete them and manually apply this unique constraint.

PR #415, opened 2024-11-06
Clean up `ArchiveService`

Swatinem:
This removes a bunch of unused functions which are not being used by `api`, as well as fix correctly passing the `is_already_gzipped` parameter.

### DIFF
--- a/shared/api_archive/archive.py
+++ b/shared/api_archive/archive.py
@@ -3,11 +3,8 @@ import logging
 from base64 import b16encode
 from enum import Enum
 from hashlib import md5
-from uuid import uuid4
 
 import sentry_sdk
-from django.conf import settings
-from django.utils import timezone
 
 from shared.api_archive.storage import StorageService
 from shared.config import get_config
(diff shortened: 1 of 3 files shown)


shared/api_archive/archive.py
```diff
@@ -33,61 +30,44 @@ class MinioEndpoints(Enum):
...
 
-    root = None
-
+    root: str
     """
-    Region where the storage is located.
+    The root level of the archive.
+    In s3 terms, this would be the name of the bucket
     """
-    region = None
```
giovanni-guidini, line 56: do we care that some self-hosted customer might be using this option or is it never used in the code?

shared/storage/base.py
```diff
@@ -46,7 +48,15 @@ def write_file(
         """
         raise NotImplementedError()
 
-    def read_file(self, bucket_name, path, file_obj=None):
+    @overload
+    def read_file(self, bucket_name: str, path: str) -> bytes: ...
```
giovanni-guidini, line 52: I didn't know about this, interesting

shared/api_archive/archive.py
giovanni-guidini: Maybe this is a setting for the underlying `StorageService`, idk

giovanni-guidini (approved):

shared/api_archive/archive.py
Swatinem: I haven’t found any code that actually references this, thats why I removed it.

PR #418, opened 2024-11-11
fix: bitbucket list_repos

giovanni-guidini:
We are noticing non-deterministic results when syncing repos for Bitbucket. The teams come from a `set`, so the evaluation order is not the same.

If you look closely the `page` was not being reset if we fail to get repos for a team, so the subsequent team (if any) would start at least from page 1, and that would cause us to not find some repos.

By starting every team search in page 0 we should have that fixed.


drazisil-codecov (approved):
Assuming tests still pass :)

Swatinem (dismissed):

giovanni-guidini:
I assumed that the pagination starts at 0 not 1, but looking over the docs there's no real indication of that... I'll actually reproduce before claiming this is the case :E

Swatinem (approved):

PR #421, opened 2024-11-12
Get rid of trigger when updating branches

adrian-codecov:
Deleting trigger and original function as it's no longer needed. Explanation + investigation here https://github.com/codecov/internal-issues/issues/1018.

### Legal Boilerplate


shared/django_apps/legacy_migrations/migrations/0005_auto_20241112_2306.py
```diff
@@ -0,0 +1,20 @@
...
+    ]
+
+    operations = [
+        RiskyRunSQL(
+            """
+            DROP TRIGGER IF EXISTS branch_update ON branches;
+            DROP FUNCTION IF EXISTS branches_update();
+            """
+        ),
+    ]
```
nora-codecov, line 20: can you give this migration file a `--name`?
please add reverse_sql=migrations.RunSQL.noop so this migration can be reversed

adrian-codecov: Yup yup, done!

shared/django_apps/legacy_migrations/migrations/legacy_sql/main/triggers/branches.py
```diff
@@ -1,32 +0,0 @@
-def run_sql(schema_editor):
-    schema_editor.execute(
-        """
-        create or replace function branches_update() returns trigger as $$
-        declare _ownerid int;
-        begin
-            -- update repos cache if main branch
-            update repos
-            set updatestamp = now(),
-                cache = update_json(cache::jsonb, 'commit', get_commit_minimum(new.repoid, new.head)::jsonb)
```
Swatinem, line 10: two followup questions:
- is the `get_commit_minimum` function unused now? can we remove it?
- first time I see the `repos.cache`. what is its purpose?

Swatinem (approved):

shared/django_apps/legacy_migrations/migrations/legacy_sql/main/triggers/branches.py
adrian-codecov: a) I'm not sure, but that should be contemplated in the commit trigger investigation
b) A field that got removed a while back, https://github.com/codecov/shared/blob/ac088a009c0c5cb7407752a1a9bcb7f3ccad1764/shared/django_apps/core/migrations/0034_remove_repository_cache.py#L20

PR #422, opened 2024-11-13
Remove pulls trigger

nora-codecov:
Shifted logic from legacy trigger to `.save()` method on the model.

This trigger may be contributing to query execution time because it runs every time any field on a `Pull` is updated. I don't think it was adding additional calls to the db (I think it was adding to the `update` query not creating an additional update query).

If the pull is no longer `open` we wipe the `flare`. I believe the intention here was to save space, there's no complex checking or intent, just save space on a field that we no longer need when a pr is closed/merged, and I think we should keep this behavior.

I replaced it with the same logic in the `save` method on the model. This guarantees that it will not cause additional db queries, and it puts the computation on the django side rather than the db.

An alternate approach - we could make this a cron job that runs overnight. I don't think the clearing of `flare` *needs* to be done immediately when the pull transitions `state`, so we could have this as a cleanup task, and we could do the change in a single `bulk_update`.

https://github.com/codecov/internal-issues/issues/1029


shared/django_apps/legacy_migrations/migrations/legacy_sql/upgrades/v440.py
```diff
@@ -57,24 +57,6 @@ def run_sql(schema_editor):
...
-            new.flare = null;
-            return new;
-        end;
-        $$ language plpgsql;
-
-        create trigger pulls_before_update_drop_flare before update on pulls
-        for each row
-        when (new.state != 'open'::pull_state)
-        execute procedure pulls_drop_flare();
-
```
nora-codecov, line 77: I don't know what this chunk of code is doing. do you think things will break if I remove this section?

shared/django_apps/core/models.py
```diff
@@ -430,11 +430,17 @@ def should_write_to_storage(self) -> bool:
...
+        default_value_class=dict,
     )
 
     def save(self, *args, **kwargs):
         self.updatestamp = timezone.now()
+        if self.state != PullStates.OPEN.value and self.flare:
+            # flare is used to draw graphs, can be quite large, so dump it when it's no longer needed
+            # flare can be None or {} to indicate empty
+            # this gets set in the db as {} since that is the default value for the field
+            self.flare = None
```
Swatinem, line 443: We should be careful with this one. I believe even accessing to check `self.flare` will fetch that file from storage (if its not being fetched ahead of time already).

And then deleting that file incurs a ~200ms GCS DELETE operation.

In particular these GCS requests are slow and might cause even more problems than a trigger unsetting a field that I believe was never written to anyway since the storage migration.

As you suggested, this might require a cronjob to asynchronously clean up storage.
Another thing that comes to mind here is that we probably never removed these things from GCS ever since that storage migration. So the trigger was completely ineffective for a long time probably.

nora-codecov: thank you times a million for this insight - I wasn't understanding what was happening with the `ArchiveField` in relation to GCS storage. If this is the case, we should make this a nightly task that runs after hours. Going to close this and reopen later

Swatinem:
@nora-codecov even though the compilation with `ArchiveField` effectively postponed the `flare` cleanup to a followup, I still think its worth removing the trigger, just without the changes to change the `flare` in the `save` hook.

It might be good to follow up here with the following action items:
- Verify that the GCS-based storage is indeed being used. It depends on feature flags, etc that should be verified.
- It might also be nice to find a trace where the `flare` is being used (read or being written).

With that verified, we should be quite confident that the `flare` column in the table is effectively unused.

Unrelated to `flare` specifically, I was talking to @giovanni-guidini a while ago that all of the `ArchiveField`s have been migrated over using a data migration. So in that case it makes sense to follow through with that migration to:
1. Always go to storage regardless of the feature flag
2. Remove those feature flags completely
3. Drop the then unused database fields complete

PR #423, opened 2024-11-13
Migrate `shared` to `uv`-based tooling

Swatinem:
This PR contains a couple of changes, happy to extract some of those into separate PRs:
- Removes some deprecated packages used for forward-compatibility with python std
- Removes some other dependencies that are unused or only used once for tests
- Migrates all dependencies to `pyproject.toml`, and uses `uv.lock` as lockfile
- Updates the docker definition and some other local project files to use `uv`


michelletran-codecov (approved):
🚀

PR #424, opened 2024-11-13
remove pickle usage in redis-backed cache helper

matt-codecov:
fixes https://github.com/codecov/internal-issues/issues/961

if the value being cached is not JSON-serializable, it will simply give up on caching it rather than raise an exception. an error will be logged to GCP so we should be able to see that occurring in GCP Logs Explorer. but from local testing i couldn't find any issues

BEWARE: technically Python allows dicts with non-`str` keys to be serialized into JSON because it coerces all keys to `str`:
```python
>>> json.loads(json.dumps( {1: 'abcde'} ))
{'1': 'abcde'}
```
this cache change will log an error and skip caching to err on the safe side

### DIFF
--- a/shared/helpers/cache.py
+++ b/shared/helpers/cache.py
@@ -1,8 +1,8 @@
 import asyncio
 import base64
 import hashlib
+import json
 import logging
-import pickle
 from functools import wraps
 from typing import Any, Callable, Hashable, List, Optional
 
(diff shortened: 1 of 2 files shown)


adrian-codecov (approved):
Any way we can rollout this canary style for some hours just to confirm your suspicions about unexpected type changes?

adrian-codecov (approved):

PR #426, opened 2024-11-14
Migrate `BitbucketServer` to `oauthlib`

Swatinem:
This pretty much copied over the oauth-related code from `Bitbucket` to `BitbucketServer`, and gets rid of the `oauth2` and `tlslite-ng` dependencies.

This does a few simplifications compared to the `Bitbucket` code, namely not caring about non-json `body`s, as I believe the existing code will only ever pass JSON payloads.

---

fixes part of https://github.com/codecov/internal-issues/issues/436
prerequisite for https://github.com/codecov/shared/pull/423

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,6 @@ setup(
         "ijson==3.*",
         "minio~=7.0",
         "mmh3",
-        "oauth2",
         "oauthlib",
         "orjson",
         "prometheus-client",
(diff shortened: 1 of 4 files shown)


michelletran-codecov (approved):
I'm OK with this, but would prefer to remove this altogether if it's not being used.

Swatinem:
> I'm OK with this, but would prefer to remove this altogether if it's not being used.

I will merge as-is then, following up on conversations to completely remove this.

Summary so far is that this is indeed not enabled in SaaS, so self-hosted might be the only ones impacted by this, and it seems like BitbucketServer itself is already EOL.

michelletran-codecov (approved):

PR #427, opened 2024-11-18
Clean up unused `minio.periodic_callback_ms` field

Swatinem:


shared/validation/install.py
```diff
@@ -270,7 +270,6 @@ def check_task_config_key(field, value, error):
                     "bucket": {"type": "string"},
                     "region": {"type": "string"},
                     "expire_raw_after_n_days": {"type": "boolean"},
-                    "periodic_callback_ms": {"type": ("boolean", "integer")},
```
adrian-codecov, line 273: Did we get rid of this? I'm not  familiar w/ the field

Swatinem: I haven’t dug up any history when this was used, but it sure isn’t used today.

adrian-codecov (approved):

PR #429, opened 2024-11-19
Create migration to delete db triggers and add functionality in python

adrian-codecov:
This PR is to address the investigation found here https://github.com/codecov/internal-issues/issues/1028. Basically deletes some triggers we no longer need + adds the `get_ownerid_if_member` function to code so we can get rid of the `owner_yaml_updated` trigger. This PR shouldn't be merged until its codecov-api counterpart is in. That PR should:

- Use the `get_ownerid_if_member` fn and branch bot updates in the owner_yaml_updated trigger [here](https://github.com/codecov/codecov-api/blob/4a2bb9b8357da69f36c24c286197e2ee99a9b3b4/codecov_auth/commands/owner/interactors/set_yaml_on_owner.py#L61)
- Delete this [TODO item](https://github.com/codecov/codecov-api/blob/4a2bb9b8357da69f36c24c286197e2ee99a9b3b4/codecov_auth/admin.py#L383) as this is the script that clears that oauth_token
- I should make a ticket to get rid of the `cache` column in the owner's database

### Legal Boilerplate


shared/django_apps/legacy_migrations/migrations/0006_delete_many_owner_triggers.py
```diff
@@ -0,0 +1,23 @@
...
+        ("legacy_migrations", "0005_delete_branch_update_db_trigger"),
+    ]
+
+    operations = [
+        RiskyRunSQL(
+            """
+            DROP TRIGGER IF EXISTS owner_yaml_updated ON owners;
+            DROP TRIGGER IF EXISTS owner_cache_state_update ON owners;
+            DROP TRIGGER IF EXISTS owner_cache_state_insert ON owners;
+            DROP TRIGGER IF EXISTS owner_token_clered ON owners;
```
Swatinem, line 19: lol, love the typo that has survived all this time :-D

shared/django_apps/utils/model_utils.py
```diff
@@ -164,3 +164,24 @@ def rollout_universe_to_override_string(rollout_universe: RolloutUniverse):
...
+    It expects a service, owner username and owner id, and returns the id of an
+    owner if the record exists.
+    """
+    owner = Owner.objects.filter(
+        service=service.lower(),
+        username=owner_username,
+        organizations__contains=[owner_id],
+        private_access=True,
+    ).first()
+    return getattr(owner, "ownerid", None)
```
Swatinem, line 187: I believe django ORM has some way to only query specific fields, instead of querying everything and then just picking the id.

shared/django_apps/legacy_migrations/migrations/0006_delete_many_owner_triggers.py
adrian-codecov: "Naming is hard" - someone in the past 😂

shared/django_apps/utils/model_utils.py
adrian-codecov: Good idea! Addressed

Swatinem (approved):

nora-codecov (approved):

PR #432, opened 2024-11-21
fix: Loosen deps ranges for compat with api, worker

suejung-sentry:
Loosen deps version ranges to be compatible with that currently in `api` and `worker`. Sets deps versions here in shared as `minimum(api, worker)` version
[api requirements.txt](https://github.com/codecov/codecov-api/blob/main/requirements.txt)
[worker requirements.txt](https://github.com/codecov/worker/blob/main/requirements.txt)

Follows up to https://github.com/codecov/shared/pull/423/files

Tested by spinning up api and worker using this version of shared with `docker compose up` & checking for any build errors. Also ran `pytest` within those containers for any build errors there


suejung-sentry:
NOTE here are the errors I'm seeing:

```
api-1  |   File "/app-dir/graphql_api/types/flake_aggregates/flake_aggregates.py", line 3, in <module>
api-1  |     import polars as pl
api-1  | ModuleNotFoundError: No module named 'polars'
```

Same error when spinning up the container and when running `pytest`

uv.lock
```diff
@@ -1385,15 +1385,15 @@ wheels = [
 
 [[package]]
 name = "sentry-sdk"
-version = "2.18.0"
+version = "2.13.0"
```
Swatinem, line 1388: this is the only package thats actually being downgraded here, and I would suggest that you not pin this, but rather define a lower bound, like for the others.

The other packages are not actually being downgraded in the lockfile, but rather their minimum version is being lowered, which has an effect on api/worker pulling in `shared` and having to resolve dependency versions. So in api/worker, we *could* end up with lower dependencies.

If you are hitting actual errors with deprecated/removed API, you can pin this to a lower version using `abs<x.y.z` syntax like for `sqlalchemy` in the `pyproject.toml`.
Not quite sure whether you want to do that here in this repo, or in api/worker rather.

michelletran-codecov (approved):
I think we can upgrade sentry-sdk in the lockfile later. I relaxed the pin, so that we have the option to use a newer version for some repos.

uv.lock
suejung-sentry: Good points - thanks!
I'll bump/fix the breaking change in sentry-sdk 2.18.0 after we get to a clean slate again - https://github.com/codecov/engineering-team/issues/2994

PR #433, opened 2024-11-22
create risky index

adrian-codecov:
Create risky migration for index that's already in prod https://github.com/codecov/engineering-team/issues/2814. We seemingly added it in the past and we're now adding the migration for it.

Worth noting, django threw an error for this index name cause it exceeded 30 characters, so I just hardcoded it after the fact. I think it's a Django error and not a sql one, given the index already exists 😂 

### Legal Boilerplate


nora-codecov (approved):

nora-codecov (commented):
make sure it's reversable

nora-codecov (approved):

PR #436, opened 2024-11-25
chore: Move worker billing functionality to shared

RulaKhaled:
As part of adding the new plan work, we are moving billing and plan stuff to shared. 
This file can be found in https://github.com/codecov/worker/blob/main/services/billing.py

### Legal Boilerplate

### DIFF
--- a/shared/billing/__init__.py
+++ b/shared/billing/__init__.py
@@ -1,15 +1,25 @@
+import logging
 from enum import Enum
+
+from django.conf import settings
+from license import get_current_license
+
+log = logging.getLogger(__name__)
 
 
 class BillingPlan(Enum):
+    users_ghm = "users"
     users_monthly = "users-inappm"
     users_yearly = "users-inappy"
     users_free = "users-free"
     users_basic = "users-basic"
+    users_trial = "users-trial"
     pr_monthly = "users-pr-inappm"
     pr_yearly = "users-pr-inappy"
     enterprise_cloud_yearly = "users-enterprisey"
     enterprise_cloud_monthly = "users-enterprisem"
+    team_monthly = "users-teamm"
+    team_yearly = "users-teamy"
 
     def __init__(self, db_name):
         self.db_name = db_name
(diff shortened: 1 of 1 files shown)


tests/unit/billing/test_enum_definitions.py
```diff
@@ -1,35 +1,99 @@
...
+    @override_settings(IS_ENTERPRISE=True)
+    def test_pr_author_enterprise_plan_check(
+        self, request, dbsession, mocker, mock_configuration
+    ):
+        owner = OwnerFactory(service="github")
+        if dbsession is not None:
+            dbsession.add(owner)
+            dbsession.flush()
+
+        encrypted_license = "wxWEJyYgIcFpi6nBSyKQZQeaQ9Eqpo3SXyUomAqQOzOFjdYB3A8fFM1rm+kOt2ehy9w95AzrQqrqfxi9HJIb2zLOMOB9tSy52OykVCzFtKPBNsXU/y5pQKOfV7iI3w9CHFh3tDwSwgjg8UsMXwQPOhrpvl2GdHpwEhFdaM2O3vY7iElFgZfk5D9E7qEnp+WysQwHKxDeKLI7jWCnBCBJLDjBJRSz0H7AfU55RQDqtTrnR+rsLDHOzJ80/VxwVYhb"
```
RulaKhaled, line 31: These are used in API, if they pose a security issue it would've been flagged.

tests/unit/billing/test_enum_definitions.py
```diff
@@ -1,35 +1,99 @@
...
+
+from shared.billing import BillingPlan, is_enterprise_cloud_plan, is_pr_billing_plan
+from shared.django_apps.codecov_auth.tests.factories import OwnerFactory
+
+
+class TestBillingServiceTestCase(object):
+    @override_settings(IS_ENTERPRISE=True)
+    @pytest.fixture
+    def dbsession(self, db):
+        return db
```
Swatinem, line 12: I would prefer these to remain fn-based tests rather than class-based tests in general.
Also this specific fixture seems to only be responsible for renaming the magic fn argument?

I believe it would be clearer to just use the existing fixture, or to add the `dbsession` fixture globally so all tests can use it. I would also clean up the used fixtures a bit, though it might not be quite clear which ones are used, as pytest fixtures are often just triggering sideeffects :-(

Swatinem (dismissed):

tests/unit/billing/test_enum_definitions.py
RulaKhaled: I removed the parts that required the tests to be class-based when I cleaned up the API tests, so function-based tests now make more sense

PR #439, opened 2024-11-26
Bundle Analysis: Make a lib to fetch route by a file name

JerrySentry:
For https://github.com/codecov/engineering-team/issues/2570

Given a file path/name of a module we want to compute what route of the application this file belongs to. This class aims to make that transformation by applying custom file based routing rules for various supported plugins. Supported plugins are:
- Remix
- NextJS Webpack
- Nuxt
- SolidStart
- SvelteKit

### Legal Boilerplate

### DIFF
--- a/shared/helpers/cache.py
+++ b/shared/helpers/cache.py
@@ -18,7 +18,7 @@ DEFAULT_TTL = 120
 def attempt_json_dumps(value: Any) -> str:
     def assert_string_keys(d: dict[Any, Any]) -> None:
         for k, v in d.items():
-            if type(k) is not str:
+            if not isinstance(k, str):
                 raise TypeError(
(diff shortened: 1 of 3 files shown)


shared/helpers/cache.py
```diff
@@ -18,7 +18,7 @@
 def attempt_json_dumps(value: Any) -> str:
     def assert_string_keys(d: dict[Any, Any]) -> None:
         for k, v in d.items():
```
JerrySentry, line 20: ruff made this change.

shared/helpers/cache.py
```diff
@@ -18,7 +18,7 @@
...
-            if type(k) is not str:
+            if not isinstance(k, str):
```
JerrySentry, line 21: ruff made this chage.

PR #442, opened 2024-11-28
Clean up existing indices on `Upload` (aka `ReportSession`)

Swatinem:
This in particular cleans up the existing indices that exist in the production DB but not in django:

We drop these indices for the following reasons:
- `upload_index_id_type_number`: This matches the above state migration,
  and it looks like the index does not actually exist in the production DB.

The following indices exist in the production DB, but not in django state/migrations:
- `reports_upload_order_number_idx`:
  We never query by the `order_number` alone, so this index is likely unused.
- `reports_upload_report_id_f6b4ffae`: Queries on `report_id` should already been covered by the
   newly added index on `report_id`+`upload_type`.
- `reports_upload_report_id_upload_type_index_ccnew`:
  This seems to be a manually added variant of the `upload_report_type_idx` index and is thus duplicated.
- `reports_upload_report_id_upload_type_order_number_index`:
  This is the same as the above, except with an additional `order_number`.
  We do use it in queries, but I doubt the index pulls its weight, as the `order_number` changes quite
  frequently so the index is costly to maintain.


adrian-codecov (approved):

Swatinem:
I updated the PR to add a bunch more manually added indices to the cleanup, and after running this, the state of the production DB should be fully aligned with the django migration state.

adrian-codecov (approved):

PR #445, opened 2024-12-02
Bundle Analysis: add ability to ingest dynamic imports

JerrySentry:
- Add a DB migration to create a new table (dynamic_imports), this table is a fk mapping between chunks and assets to specify what dynamically imported assets a chunk has.
- Add a new v3 JSON stats file parser that reads chunk.dynamicImport array of strings and stores this into the dynamic_imports table

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/migrations/v004_add_dynamic_imports.py
+++ b/shared/bundle_analysis/migrations/v004_add_dynamic_imports.py
@@ -1,0 +1,27 @@
+from sqlalchemy import text
+from sqlalchemy.orm import Session
+
+
+def add_dynamic_imports(db_session: Session):
+    """
+    Adds a table called dynamic_imports (DynamicImport model name)
+    This table represents for a given Chunk what are its dynamically
+    imported Assets, if applicable.
+
+    There is no data available to migrate, any older versions of bundle
+    reports will be considered to not have dynamic imports
+    """
+    stmts = [
+        """
+        CREATE TABLE dynamic_imports (
+            chunk_id integer not null,
+            asset_id integer not null,
+            primary key (chunk_id, asset_id),
+            foreign key (chunk_id) references chunks (id),
+            foreign key (asset_id) references assets (id)
+        );
+        """,
+    ]
+
+    for stmt in stmts:
+        db_session.execute(text(stmt))
--- a/shared/bundle_analysis/parsers/v3.py
+++ b/shared/bundle_analysis/parsers/v3.py
@@ -1,0 +1,424 @@
+import json
+import logging
+import re
+import uuid
+from collections import defaultdict
+from typing import List, Tuple
+
+import ijson
+import sentry_sdk
+from sqlalchemy.orm import Session as DbSession
+from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
+
+from shared.bundle_analysis.models import (
+    Asset,
+    AssetType,
+    Bundle,
+    Chunk,
+    DynamicImport,
+    Module,
+    Session,
+    assets_chunks,
+    chunks_modules,
+)
+from shared.bundle_analysis.parsers.base import ParserTrait
+from shared.bundle_analysis.utils import get_extension
+
+log = logging.getLogger(__name__)
+
+
+"""
+Version 3 Schema
+{
+    "version": "3",
+    "plugin": {
+      "name": str
+      "version": str
+    },
+    "builtAt": int,
+    "duration": int,
+    "bundler": { "name": str, "version": str },
+    "bundleName": str,
+    "assets": [{
+        "name": str",
+        "size": int,
+        "gzipSize": int,
+        "normalized": str
+    }],
+    "chunks": [{
+        "id": str,
+        "uniqueId": str,
+        "entry": bool,
+        "initial": bool,
+        "files": [str],
+        "names": [str],
+        "dynamicImports": [str]
+    }],
+    "modules": [{
+        "name": str,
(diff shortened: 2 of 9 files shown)


shared/bundle_analysis/parsers/v3.py
```diff
@@ -0,0 +1,424 @@
+import json
```
JerrySentry, line 1: This entire file is basically a carbon copy of v2, except what is identified below. Note that each time a new stats file json schema  is updated we create a new parser version 95% of the code between each version would be the same.

shared/bundle_analysis/parsers/v3.py
```diff
@@ -0,0 +1,424 @@
...
+        "normalized": str
+    }],
+    "chunks": [{
+        "id": str,
+        "uniqueId": str,
+        "entry": bool,
+        "initial": bool,
+        "files": [str],
+        "names": [str],
+        "dynamicImports": [str]
```
JerrySentry, line 55: This is the only change to the schema, also the `"version": "3"`

shared/bundle_analysis/parsers/v3.py
```diff
@@ -0,0 +1,424 @@
...
+                )
+            )
+
+            self.module_chunk_unique_external_ids_index[self.module.name] = (
+                self.module_chunk_unique_external_ids
+            )
+            # reset parser state
+            self.module = None
+            self.module_chunk_unique_external_ids = []
+
```
JerrySentry, line 324: This function below is a new one, it collects all the pairs of (chunk, asset) that is a dynamic import to be saved into the DB at the end.

shared/bundle_analysis/parsers/v3.py
```diff
@@ -0,0 +1,424 @@
...
+                # Needs to use ORM-style insert to update the models since they
+                # will be used later in dynamic import processing
+                self.db_session.add_all(self.chunk_list)
+
+                if self.module_list:
+                    insert_modules = Module.__table__.insert().values(self.module_list)
+                    self.db_session.execute(insert_modules)
+
+                self.db_session.flush()
+
```
JerrySentry, line 158: Below is where we have the list of items to insert into the DB. Also a small above change
```
self.db_session.add_all(self.chunk_list)
```
compared to before of `Chunk.__table__.insert().values`

shared/bundle_analysis/parsers/v3.py
```diff
@@ -0,0 +1,424 @@
...
+        self.chunk_asset_names = []
+        self.module = None
+        self.module_chunk_unique_external_ids = []
+
+        self.asset_list = []
+        self.chunk_list = []
+        self.module_list = []
+
+        # dynamic imports: mapping between Chunk and each file name of its dynamic imports
+        self.dynamic_imports_mapping = defaultdict(
```
suejung-sentry, line 102: Thoughts on naming this `dynamic_import_file_names_by_chunk` or similar?

shared/bundle_analysis/parsers/v3.py
```diff
@@ -0,0 +1,424 @@
...
+                [
+                    dict(
+                        chunk_id=chunk_unique_id_to_id[unique_external_id],
+                        module_id=module_id,
+                    )
+                    for unique_external_id in chunk_unique_external_ids
+                ]
+            )
+        if inserts:
+            self.db_session.execute(chunks_modules.insert(), inserts)
```
suejung-sentry, line 424: Do we need any transaction safety/rollback on the `assets_chunks.insert()` in above and this `chunks_modules.insert()` if one succeeds but the next fails?

shared/bundle_analysis/parsers/v3.py
```diff
@@ -0,0 +1,424 @@
...
+
+            self.module_chunk_unique_external_ids_index[self.module.name] = (
+                self.module_chunk_unique_external_ids
+            )
+            # reset parser state
+            self.module = None
+            self.module_chunk_unique_external_ids = []
+
+    def _parse_dynamic_imports(self) -> List[dict]:
+        dynamic_imports_list = []
```
suejung-sentry, line 326: May be nice to have a more specific type on this `dynamic_imports_list` and then we can omit the type from the var name. Can see that being a pain though

suejung-sentry (approved):
Can't guarantee I fully tracked all the mappings across name, path, id, unique_id, unique_external_id to catch any bugs in there but looks like you've got some great test fixtures so looks good to me!

shared/bundle_analysis/migrations/v004_add_dynamic_imports.py
```diff
@@ -0,0 +1,27 @@
...
+    reports will be considered to not have dynamic imports
+    """
+    stmts = [
+        """
+        CREATE TABLE dynamic_imports (
+            chunk_id integer not null,
+            asset_id integer not null,
+            primary key (chunk_id, asset_id),
+            foreign key (chunk_id) references chunks (id),
+            foreign key (asset_id) references assets (id)
```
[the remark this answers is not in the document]
JerrySentry: The good thing about these "databases" are that it is super small (like 100kB), it only matches bundle data for 1 singular commit. This means we don't need to worry about performance on querying things haha.

shared/bundle_analysis/parsers/v3.py
JerrySentry: This entire `parse` should be already in a transaction, we don't actually call `session.commit` in here which writes everything into the filesystem, that happens on the caller of the parser.

PR #447, opened 2024-12-04
Bundle Analysis: Routing report and comparisons

JerrySentry:
- Adds a BundleRouteReport to retrieve total size of each route to enable comparisons
- Adds a BundleRoutesComparison to do comparisons between base and head BundleRouteReports. This will be used by PR comments.

closes: https://github.com/codecov/engineering-team/issues/3032

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/comparison.py
+++ b/shared/bundle_analysis/comparison.py
@@ -3,7 +3,7 @@ from collections import defaultdict
 from dataclasses import dataclass
 from enum import Enum
 from functools import cached_property
-from typing import Iterator, List, MutableSet, Optional, Tuple
+from typing import Dict, Iterator, List, MutableSet, Optional, Tuple
 
 import sentry_sdk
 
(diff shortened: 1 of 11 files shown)


shared/bundle_analysis/comparison.py
```diff
@@ -176,6 +188,68 @@ def _match_assets(
...
+        If a route exists on head but not base that is considered "added" and +100% percentage delta
+        Otherwise it is considered "changed" and percentage delta = (diff_size / base_size) * 100
+        """
+        base_sizes = self.base_report.get_size()
+        head_sizes = self.head_report.get_size()
+
+        all_routes, results = base_sizes.keys() | head_sizes.keys(), []
+        for route_name in all_routes:
+            # Added new route
+            if route_name not in base_sizes:
```
suejung-sentry, line 218: Do we need to add the case where base_sizes[route_name] is 0 to avoid a potential divide by 0 issue at line 240? Can 0 even happen - maybe with rounding?

suejung-sentry (approved):
cool stuff!

PR #449, opened 2024-12-05
Bundle Analysis: fix routes and add sizes to RouteChange

JerrySentry:
Will need the absolute sizes for Route in PR comments

### Legal Boilerplate


shared/bundle_analysis/report.py
```diff
@@ -304,8 +304,10 @@ def routes(self) -> Dict[str, List[AssetReport]]:
         """
         route_map = defaultdict(list)
         for asset_report in self.asset_reports():
-            for route in asset_report.routes():
-                route_map[route].append(asset_report)
```
JerrySentry, line 308: fixes a bug where routes is null

suejung-sentry (approved):

PR #450, opened 2024-12-05
Remove pulls trigger

nora-codecov:
Part of the quest to reduce locks by eliminating triggers https://github.com/codecov/internal-issues/issues/1029

The trigger is deleted by these changes. The behavior is replaced in a cron job on worker https://github.com/codecov/worker/pull/947.

What the trigger did: every time a pull was updated, if it was transitioning from `OPEN` to any other `state`, it set the `flare` field to `null`.
`flare` is an Archive field, so the value is either in `_flare` on the object or in our Archive storage.

Replacement behavior: see `FlareCleanupTask` https://github.com/codecov/worker/pull/947
Once the trigger is removed, when the pull transitions from `OPEN` to any other `state`, leave `flare` alone in order to reduce locks and wait time.
The cron job runs overnight, gets all pulls that are non-`OPEN` and have flare (either in our db or in Archive storage), clears it (either in our db or in Archive storage) so `flare` is `null`.

more about `flare`: 
- see https://github.com/codecov/codecov-api/pull/999 and my summary of the findings in https://github.com/codecov/internal-issues/issues/1029


shared/django_apps/core/models.py
```diff
@@ -418,6 +418,19 @@ def external_id(self):
...
+        The majority of flare graphs are used in pr comments, so we keep the (maybe large) flare available while
+        the pull is OPEN. If the pull is not OPEN, we dump the flare to save space. We will recreate it on
+        the fly if it is needed for a non-OPEN pull (see GraphHandler view in api).
+        Flare cleanup is handled by FlareCleanupTask in worker.
+        """
+        if self.state != PullStates.OPEN.value:
+            # while a pull is OPEN, we check whether to write_to_storage
+            # when a pull is no longer OPEN, we no longer want the flare in storage.
+            # The nightly cron job cleans up the value in storage (see FlareCleanupTask in worker)
+            return False
```
Swatinem, line 433: this would rather mean that the flare is being stored in the JSON field of the table.
I’m unsure what would happen if you change this value based on the `state`. will is copy over data from storage to the database? or will that only happen if you save a new value in the field?

Swatinem (approved):
I’m unsure related to the `should_write_to_storage` changes. I don’t think those are needed necessarily?

shared/django_apps/core/models.py
nora-codecov: good call - I think I was thinking that if this came back `False` it wouldn't be saved anywhere, but you're right, this would just force it onto the db field. Removing.

Swatinem (approved):

shared/storage/aws.py
```diff
@@ -147,7 +147,7 @@ def delete_file(self, bucket_name, path):
         except ClientError:
             raise
 
-    def delete_files(self, bucket_name, paths=[]):
```
nora-codecov, line 150: This default value is dangerous and unnecessary. fixed the occurrences I found across the project.

shared/api_archive/archive.py
```diff
@@ -64,7 +64,10 @@ def __init__(self, repository, ttl=None):
         self.ttl = ttl or int(get_config("services", "minio", "ttl", default=self.ttl))
 
         self.storage = StorageService()
-        self.storage_hash = self.get_archive_hash(repository)
+        if repository:
+            self.storage_hash = self.get_archive_hash(repository)
+        else:
+            self.storage_hash = None
```
nora-codecov, line 70: Making ArchiveService dependent on a repository was complicating my flare cleanup task.
It seems like repository is only needed to make the hash for reading and writing to the archive - you don't need it to delete.
To make my flare cleanup task much lighter, I removed the requirement for a repository in order to use ArchiveService. I added checks to the ArchiveStorage functions where the hash is used, so that you can't do something like write to storage without a hash.

PR #452, opened 2024-12-09
Add conditional based on upload type

adrian-codecov:
The copilot AI button appears in every PR comment but should not appear in the BA PR comment. This solution adds a way to specify the upload type and manipulate the data being sent to GH.

### Legal Boilerplate

### DIFF
--- a/shared/typings/torngit.py
+++ b/shared/typings/torngit.py
@@ -1,5 +1,10 @@
+from enum import Enum
 from typing import Dict, List, Optional, TypedDict, Union
 
+class UploadType(Enum):
+    COVERAGE = "coverage"
+    TEST_RESULTS = "test_results"
+    BUNDLE_ANALYSIS = "bundle_analysis"
 
 class OwnerInfo(TypedDict):
     service_id: str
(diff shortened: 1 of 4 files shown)


shared/typings/torngit.py
```diff
@@ -25,9 +30,13 @@ class GithubInstallationInfo(TypedDict):
     app_id: Optional[int]
     pem_path: Optional[str]
 
+class AdditionalData(TypedDict):
+    upload_type: Optional[UploadType]
```
giovanni-guidini, line 34: When you say "Optional" this indicates that the key will always be there.
Looking above the initialization is just `{}` (no key).

If the key may or may not be present then use `NotRequired` (https://docs.python.org/3/whatsnew/3.11.html#pep-655-marking-individual-typeddict-items-as-required-or-not-required)

I'm sure other fields in this file suffer from the same problem 😅

giovanni-guidini (approved):
Interesting choice to extend the whole `TorngitInstanceData` instead of just the function. I suppose we can benefit from that additional_data tho.

I was surprised we didn't have an `UploadType` enum already.

shared/typings/torngit.py
adrian-codecov: We have an `UploadType` that's the DB Enum. We do have a ReportType in Worker that has what I'm looking for, so I'll add that to this change

And will add the `NotRequired`, TIL

PR #453, opened 2024-12-10
feat: Add new properties for is____Plan on Plan Service

ajay-sentry:
Needed for https://github.com/codecov/codecov-api/pull/1039, This PR adds the properties being fetched by the resolvers

Related Ticket -- https://github.com/codecov/engineering-team/issues/2996

### Legal Boilerplate

### DIFF
--- a/shared/plan/service.py
+++ b/shared/plan/service.py
@@ -316,39 +316,27 @@ class PlanService:
 
     @property
     def is_enterprise_plan(self) -> bool:
-        if self.plan_name in ENTERPRISE_CLOUD_USER_PLAN_REPRESENTATIONS:
-            return True
-        return False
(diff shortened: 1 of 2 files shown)


shared/plan/service.py
```diff
@@ -311,3 +313,42 @@ def has_seats_left(self) -> bool:
             self.plan_activated_users is None
             or len(self.plan_activated_users) < self.plan_user_count
         )
+
+    @property
+    def is_enterprise_plan(self) -> bool:
```
JerrySentry, line 318: nit: can just do `return self.plan_name in ...` for all these

JerrySentry (approved):

shared/plan/service.py
ajay-sentry: good point, will update

PR #459, opened 2024-12-20
Ensure that language validation works on all supported languages

michelletran-codecov:
The database has a set of enum validation which is different from the in-code validation that we do when checking that the language is valid. This should ensure that the language set looks the same.

Resolves this issue: https://github.com/codecov/engineering-team/issues/2061

Might want to try this in worker/staging before merging because it uses a Django property.
### Legal Boilerplate

### DIFF
--- a/shared/torngit/base.py
+++ b/shared/torngit/base.py
@@ -4,6 +4,7 @@ from typing import Dict, List, Optional, Tuple
 
 import httpx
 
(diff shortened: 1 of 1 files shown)


adrian-codecov (approved):
Ooooh interesting, yeah lets give it a shot in staging and see how it goes! Great catch

PR #461, opened 2025-01-03
PlanService updates for GitLab

nora-codecov:
update PlanService so GitLab orgs use the root org's plan, fold is_pr_billing_plan into PlanService

IF an OwnerOrg is a GitLab subgroup OR has an Account, the `plan` on the OwnerOrg is not the one we want to use! We want the `plan` on the root Org or Account object instead.

PlanService already had the smarts to check for an `Account` and get the Account `plan`, but now it also has the smarts to check whether a GitLab OwnerOrg has a parent, then use the `plan` from the root Org. 

Part of https://github.com/codecov/engineering-team/issues/2710

### DIFF
--- a/shared/billing/__init__.py
+++ b/shared/billing/__init__.py
@@ -1,23 +1,28 @@
 from enum import Enum
 
 from django.conf import settings
+from typing_extensions import deprecated
 
 from shared.license import get_current_license
+from shared.plan.constants import PlanName
 
 
+@deprecated("Use PlanService")
 class BillingPlan(Enum):
-    users_ghm = "users"
-    users_monthly = "users-inappm"  # not pr_billing_plan
-    users_yearly = "users-inappy"  # not pr_billing_plan
-    users_free = "users-free"
-    users_basic = "users-basic"
-    users_trial = "users-trial"
-    pr_monthly = "users-pr-inappm"
-    pr_yearly = "users-pr-inappy"
-    enterprise_cloud_yearly = "users-enterprisey"
-    enterprise_cloud_monthly = "users-enterprisem"
-    team_monthly = "users-teamm"
-    team_yearly = "users-teamy"
+    users_basic = PlanName.BASIC_PLAN_NAME.value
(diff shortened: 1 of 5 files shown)


shared/billing/__init__.py
```diff
@@ -37,6 +37,7 @@ def is_enterprise_cloud_plan(plan: BillingPlan) -> bool:
 
 
 def is_pr_billing_plan(plan: str) -> bool:
+    # use is_pr_billing_plan() in PlanService instead of accessing this directly
```
giovanni-guidini, line 40: Maybe it's worth to mark it as [deprecated](https://typing.readthedocs.io/en/latest/spec/directives.html#deprecated)?

shared/plan/service.py
```diff
@@ -46,7 +47,14 @@ def __init__(self, current_org: Owner):
         Raises:
             ValueError: If the organization's plan is unsupported.
         """
-        self.current_org = current_org
+        if (
+            current_org.service == Service.GITLAB.value
+            and current_org.parent_service_id
```
giovanni-guidini, line 52: This doesn't seem correct... you are checking for `current_org.parent_service_id` not being `None`, but `self.current_org` is getting the value of `current_org.root_organization`

shared/plan/constants.py
```diff
@@ -29,8 +29,8 @@ class PlanName(enum.Enum):
     TRIAL_PLAN_NAME = "users-trial"
     CODECOV_PRO_MONTHLY = "users-pr-inappm"
     CODECOV_PRO_YEARLY = "users-pr-inappy"
-    SENTRY_MONTHLY = "users-sentrym"
-    SENTRY_YEARLY = "users-sentryy"
+    SENTRY_MONTHLY = "users-sentrym"  # not in BillingPlan
+    SENTRY_YEARLY = "users-sentryy"  # not in BillingPlan
```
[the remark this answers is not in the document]
nora-codecov: Talked with Ajay, this is an oversight, adding to BillingPlan

giovanni-guidini (approved):

PR #464, opened 2025-01-06
[Bundle Analysis] Cache config service get_or_create

JerrySentry:
Add get_or_create, will have to use this instead of update_or_create in the BA processor. Currently we will always overwrite what the caching bool to true when doing a new commit into default branch. When switching to this the intention is that if the bundle cache data already exists we will not overwrite it to true.

### Legal Boilerplate

### DIFF
--- a/shared/django_apps/bundle_analysis/service/bundle_analysis.py
+++ b/shared/django_apps/bundle_analysis/service/bundle_analysis.py
@@ -9,7 +9,7 @@ class BundleAnalysisCacheConfigService:
         )
 
     @staticmethod
-    def create_cache_option(repo_id, name, is_caching=True) -> None:
+    def create_if_not_exists(repo_id, name, is_caching=True) -> None:
         CacheConfig.objects.get_or_create(
             repo_id=repo_id, bundle_name=name, defaults={"is_caching": is_caching}
         )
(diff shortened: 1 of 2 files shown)


shared/django_apps/bundle_analysis/service/bundle_analysis.py
```diff
@@ -7,3 +7,9 @@ def update_cache_option(repo_id, name, is_caching=True) -> None:
         CacheConfig.objects.update_or_create(
             repo_id=repo_id, bundle_name=name, defaults={"is_caching": is_caching}
         )
+
+    @staticmethod
+    def create_cache_option(repo_id, name, is_caching=True) -> None:
```
suejung-sentry, line 12: any thoughts on naming the method same with `get_or_create` or `create_if_not_exists`?
also, wondering if that means we will remove the `update_or_create` one now?

suejung-sentry (approved):

shared/django_apps/bundle_analysis/service/bundle_analysis.py
JerrySentry: I like `create_if_not_exists`, `get_or_create` implies I need to return something in this method.
`update_or_create` will be used by the API in the next couple of PRs, but I'm gonna rename it to `update_cache_option` because it won't be creating.

PR #467, opened 2025-01-09
feat: Create plans and tiers tables

RulaKhaled:
To replace plans representation, we're making plans configurable through the DB. 

### Legal Boilerplate


shared/django_apps/codecov_auth/models.py
```diff
@@ -997,3 +997,50 @@ def save(self, *args, **kwargs):
                 id=self.id
             ).update(is_active=False)
         return super().save(*args, **kwargs)
+
+
+class BillingRate(models.TextChoices):
+    MONTHLY = "monthly"
+    YEARLY = "yearly"
```
nora-codecov, line 1004: why not use `PlanBillingRate`?

shared/django_apps/codecov_auth/models.py
```diff
@@ -997,3 +997,50 @@ def save(self, *args, **kwargs):
...
+
+
+class Plans(BaseModel):
+    tier = models.ForeignKey(
+        "Tiers", on_delete=models.CASCADE, related_name="plans", db_index=True
+    )
+    base_unit_price = models.IntegerField(default=0)
```
nora-codecov, line 1011:
```suggestion
    base_unit_price = models.IntegerField(default=0, blank=True)
```

shared/django_apps/codecov_auth/models.py
```diff
@@ -997,3 +997,50 @@ def save(self, *args, **kwargs):
...
+    )
+    is_active = models.BooleanField(default=True)
+    marketing_name = models.CharField(max_length=255)
+    max_seats = models.IntegerField(null=True, blank=True)
+    monthly_uploads_limit = models.IntegerField(null=True, blank=True)
+    paid_plan = models.BooleanField(default=False)
+    plan_name = models.CharField(max_length=255, unique=True)
+
+    class Meta:
+        db_table = "plans"
```
nora-codecov, line 1026: is this needed?

shared/django_apps/codecov_auth/models.py
```diff
@@ -997,3 +997,50 @@ def save(self, *args, **kwargs):
...
+class Tiers(BaseModel):
+    tier_name = models.CharField(max_length=255, unique=True)
+    bundle_analysis = models.BooleanField(default=False)
+    test_analytics = models.BooleanField(default=False)
+    flaky_test_detection = models.BooleanField(default=False)
+    patch_coverage = models.BooleanField(default=False)
+    private_repo_support = models.BooleanField(default=False)
+
+    class Meta:
+        db_table = "tiers"
```
nora-codecov, line 1042: prob not needed for a new table unless there's naming overlap with another table

nora-codecov (commented):
looking good! left some comments, happy to discuss any if you want :)

I think @ajay-sentry said that there will be a lot of django admin use of these tables - I would make those pages now and test them to make sure your fields are the way you want

shared/django_apps/codecov_auth/models.py
```diff
@@ -997,3 +997,50 @@ def save(self, *args, **kwargs):
...
+
+class BillingRate(models.TextChoices):
+    MONTHLY = "monthly"
+    YEARLY = "yearly"
+
+
+class Plans(BaseModel):
+    tier = models.ForeignKey(
+        "Tiers", on_delete=models.CASCADE, related_name="plans", db_index=True
+    )
```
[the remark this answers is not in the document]
ajay-sentry: Agreed with Nora, probably don't need an index here

shared/django_apps/codecov_auth/models.py
ajay-sentry: I think this one should always have a value, assuming blank=True means nullable

shared/django_apps/codecov_auth/models.py
```diff
@@ -997,3 +997,50 @@ def save(self, *args, **kwargs):
...
+        choices=BillingRate.choices,
+        null=True,
+        blank=True,
+    )
+    is_active = models.BooleanField(default=True)
+    marketing_name = models.CharField(max_length=255)
+    max_seats = models.IntegerField(null=True, blank=True)
+    monthly_uploads_limit = models.IntegerField(null=True, blank=True)
+    paid_plan = models.BooleanField(default=False)
+    plan_name = models.CharField(max_length=255, unique=True)
```
ajay-sentry, line 1023: just a thought here, should we just call this "name?" since it's already on the plan 🤔

shared/django_apps/codecov_auth/models.py
ajay-sentry: what does this do? :O

shared/django_apps/codecov_auth/models.py
```diff
@@ -997,3 +997,50 @@ def save(self, *args, **kwargs):
...
+    def __str__(self):
+        return self.plan_name
+
+
+class Tiers(BaseModel):
+    tier_name = models.CharField(max_length=255, unique=True)
+    bundle_analysis = models.BooleanField(default=False)
+    test_analytics = models.BooleanField(default=False)
+    flaky_test_detection = models.BooleanField(default=False)
+    patch_coverage = models.BooleanField(default=False)
```
ajay-sentry, line 1038: This should be project coverage right?

shared/django_apps/codecov_auth/models.py
ajay-sentry: this should be annually instead of yearly as well I think https://github.com/codecov/gazebo/blob/113a97f9e6a40ab9162576599308e123a25626aa/src/shared/utils/billing.ts#L29

RulaKhaled: didn't know we had it already

RulaKhaled: didn't know we have it but i think we can't use enum.Enum directly in the choices argument of a Django model field, we'll have to do `[(tag.value, tag.name) for tag in PlanBillingRate]` which is not very clean.

I'll update to annually

shared/django_apps/codecov_auth/models.py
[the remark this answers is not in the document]
RulaKhaled: makes sense, i always like to use indexes you never know how big the table will grow to be

shared/django_apps/codecov_auth/models.py
RulaKhaled: blank true means it can be null in the DB but it can't be optional in a form

shared/django_apps/codecov_auth/models.py
RulaKhaled: it's just a clean way of naming the database

shared/django_apps/codecov_auth/models.py
RulaKhaled: yes good catch

shared/django_apps/codecov_auth/tests/factories.py
```diff
@@ -169,3 +171,31 @@ class Meta:
...
+class TierFactory(DjangoModelFactory):
+    class Meta:
+        model = Tiers
+
+    tier_name = factory.Faker("name")
+    bundle_analysis = False
+    test_analytics = False
+    flaky_test_detection = False
+    project_coverage = False
+    private_repo_support = False
```
nora-codecov, line 185: if you had `blank=True` on the bool fields on the model, you wouldn't need to name any of these fields, they would automatically be `False` when you make an obj with the factory

ajay-sentry: does it default to False or None in that case?

ajay-sentry: This stack overflow page seems to give a mixed view -- https://stackoverflow.com/questions/8609192/what-is-the-difference-between-null-true-and-blank-true-in-django

ajay-sentry: I realize now that we already had default=False on the model too 🤦‍♂️

ajay-sentry (approved):

nora-codecov (approved):

PR #468, opened 2025-01-09
Add the `CompareCommit` models to Django

Swatinem:


Swatinem:
Going with #469 instead

PR #469, opened 2025-01-09
Copy over some django_apps and their associated models to `shared`

Swatinem:
This just straight-up copies the following directories from `codecov-api`:

`compare`, `labelanalysis`, `profiling` and `staticanalysis`

I would like to have access to these models from the worker, so thats why I am moving those to `shared`.


adrian-codecov (approved):
This overall lgtm! When reimporting, can we try letting it sit in staging for a day or something just to make sure they're working fine?

PR #474, opened 2025-01-14
feat: create test_analytics app and database

joseph-sentry:
- creates a new test_analytics django app
- modifies the db_settings to declare a new test analytics database
- modifies the db router to route the models from the new django app to the new database


shared/django_apps/db_routers/__init__.py
```diff
@@ -12,46 +12,55 @@ class MultiDatabaseRouter:
...
-            else:
-                return "default"
+        match model._meta.app_label:
+            case "timeseries":
+                if settings.TIMESERIES_DATABASE_READ_REPLICA_ENABLED:
+                    return "timeseries_read"
+                else:
+                    return "timeseries"
+            case "test_analytics":
+                return "test_analytics"
```
adrian-codecov, line 22: Unlike timeseries, there's not read/write instances right

adrian-codecov (approved):

Swatinem (approved):
The intention here is that this will eventually replace the `Flake` thats defined in `reports` right now?
How about all the other models related to TA?

joseph-sentry:
> The intention here is that this will eventually replace the Flake thats defined in reports right now?
How about all the other models related to TA?

I don't think we'll be needing them anymore given we'll be reading from BQ. The actual win in creating a new app and database connection is that we can scale the database for TA separately from coverage and all our other products

PR #477, opened 2025-01-15
Add USE_MINIO feature flag

joseph-sentry:
we want to start by rolling the worker out to use the MinioStorageService for only our repos, then incrementally roll it out to more users


Swatinem (approved):
lets do it 💪🏻

PR #479, opened 2025-01-16
Shared: Migrate to Plan / Tier Tables

ajay-sentry:
### Description

This PR aims to migrate off the plan consts in shared in favor of the Plan data model. The integration can be summed up as doing DB calls throughout where previously we were referencing a constant value. 

Typically a query for a plan will use the plan name (identified as "plan") on the owner model to first retrieve a user's plan. At the same time, if the tier is needed, we will also prefetch that object as well to streamline the number of DB queries we have. 

As much as possible we default to either using the plan "cached" from a higher level query, as in the Plan Service; or we build up a query and execute in the case of availablePlans. Overall this will result in slightly higher load on our database, if there are any charts and/or metrics anyone has readily available with those higher level monitoring specs I'd love to take a look.

We will look to clean up those consts and any lingering references (only in tests at this point) in an upcoming PR.

This PR relates to the API/Worker versions here:
- https://github.com/codecov/worker/pull/1007
- https://github.com/codecov/codecov-api/pull/1099

Notion with some more info about the motivation and where we are in the process: https://www.notion.so/sentry/Streamlining-Plan-Creation-with-the-Plan-Configuration-Table-1428b10e4b5d808ca5b2d4f5b66493a1

Closes https://github.com/codecov/engineering-team/issues/3251

### Legal Boilerplate


shared/plan/constants.py
```diff
@@ -109,6 +111,25 @@ def convert_to_DTO(self) -> dict:
...
+def convert_to_DTO(plan) -> dict:
+    return {
+        "marketing_name": plan.marketing_name,
+        "value": plan.name,
+        "billing_rate": plan.billing_rate,
+        "base_unit_price": plan.base_unit_price,
+        "benefits": plan.benefits,
+        "tier_name": plan.tier.tier_name,
+        "monthly_uploads_limit": plan.monthly_uploads_limit,
+        "trial_days": TrialDaysAmount.CODECOV_SENTRY.value,
```
ajay-sentry, line 123: this should still be 0 for a number of plans, will make the change

ajay-sentry: turns out we don't need this at all, removing from DTO

tests/unit/plan/test_plan.py
```diff
@@ -16,15 +20,173 @@
...
     PlanName,
+    PlanPrice,
+    TierName,
     TrialDaysAmount,
     TrialStatus,
 )
 from shared.plan.service import PlanService
 
 
+def mock_all_plans_and_tiers():
```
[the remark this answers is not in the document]
RulaKhaled: Yes! A good place for it could be the plan constants. I think the API won't be able to pull from tests helpers; we would need it there; otherwise we could just create our own mocks in API

shared/django_apps/codecov_auth/models.py
```diff
@@ -221,10 +220,22 @@ def pretty_plan(self) -> dict | None:
         This is how we represent the details of a plan to a user, see plan.constants.py
         We inject quantity to make plan management easier on api, see PlanSerializer
         """
-        if self.plan in USER_PLAN_REPRESENTATIONS:
-            plan_details = asdict(USER_PLAN_REPRESENTATIONS[self.plan])
-            plan_details.update({"quantity": self.plan_seat_count})
-            return plan_details
+        plan_details = Plan.objects.get(name=self.plan)
+        if plan_details:
+            return {
```
adrian-codecov, line 225: Could we make this a class method, like serialized plan or something

``` python
class Plan(BaseModel):
...

  def serialized_plan(self): -> <you could also type this into another dataclass or something if you want brownie points)
   return {...} # this is where you'd put this "pretty plan"

```

shared/plan/constants.py
```diff
@@ -109,6 +110,24 @@ def convert_to_DTO(self) -> dict:
         }
 
 
+def convert_to_DTO(plan) -> dict:
```
adrian-codecov, line 113: Thoughts on typing this?

shared/plan/service.py
```diff
@@ -55,19 +47,20 @@ def __init__(self, current_org: Owner):
             self.current_org = current_org.root_organization
         else:
             self.current_org = current_org
-        if self.current_org.plan not in USER_PLAN_REPRESENTATIONS:
+
+        if not Plan.objects.filter(name=self.current_org.plan).exists():
```
adrian-codecov, line 51: Remind me, are we porting all plans, including "obsolete" plans like the super super old ones that we don't use? If so, changing those "if name not in USER_PLAN_REPRESENTATION" to this wouldn't be 100% equivalent

shared/plan/service.py
```diff
@@ -89,25 +82,25 @@ def has_account(self) -> bool:
         return self.current_org.account is not None
 
     @property
-    def plan_data(self) -> PlanData:
+    def plan_data(self) -> Plan:
```
adrian-codecov, line 85: Have you confirmed this implementation effectively acts as cacheing the field? Since this is now a db query, we'd want this to be cached - perhaps look into cached_property decorator

shared/plan/service.py
```diff
@@ -159,29 +152,35 @@ def monthly_uploads_limit(self) -> Optional[int]:
         return self.plan_data.monthly_uploads_limit
 
     @property
-    def tier_name(self) -> str:
+    def tier_name(self) -> TierName:
         """Returns the tier name of the plan."""
-        return self.plan_data.tier_name
+        return self.plan_data.tier.tier_name
```
adrian-codecov, line 157: Since this has a relationship to tier, might be worth in the query of plan_data to prefetch this relationship, otherwise you'll have an n+1

shared/plan/service.py
```diff
@@ -159,29 +152,35 @@ def monthly_uploads_limit(self) -> Optional[int]:
...
 
-    def available_plans(self, owner: Owner) -> List[PlanData]:
+    def available_plans(self, owner: Owner) -> List[Plan]:
         """Returns the available plans for the owner and organization."""
-        available_plans = [BASIC_PLAN]
+        available_plans = {Plan.objects.get(name=PlanName.BASIC_PLAN_NAME.value)}
 
-        if self.plan_name == FREE_PLAN.value:
-            available_plans.append(FREE_PLAN)
+        curr_plan = Plan.objects.get(name=self.plan_name)
```
adrian-codecov, line 163: Should this be the same as `plan_data`? If not, how does the has_account affect this logic?

shared/plan/constants.py
ajay-sentry: I was getting a circular dependency trying to import the Plan class and type it that way, is there another way we can do it?

shared/plan/service.py
```diff
@@ -159,29 +152,35 @@ def monthly_uploads_limit(self) -> Optional[int]:
...
-            self.plan_activated_users is None
+            not self.plan_activated_users
             or len(self.plan_activated_users) <= TEAM_PLAN_MAX_USERS
         ):
-            available_plans += TEAM_PLAN_REPRESENTATIONS.values()
+            available_tiers.append(TierName.TEAM.value)
 
-        return [plan.convert_to_DTO() for plan in available_plans]
+        available_plans.update(
+            Plan.objects.filter(tier__tier_name__in=available_tiers, is_active=True)
```
adrian-codecov, line 180: If we're going to always do this at least for the PRO value, would it make sense for the first available_plans to include that parameter, and if available_tiers doesn't change, you save yourself 1 query every time

I'd also look into cacheing this one for perf

shared/plan/service.py
```diff
@@ -236,10 +235,12 @@ def start_trial_manually(self, current_owner: Owner, end_date: datetime) -> None
             No value
         """
         # Start a new trial plan for free users currently not on trial
-        if self.plan_name in FREE_PLAN_REPRESENTATIONS:
+
+        plan = Plan.objects.get(name=self.plan_name)
```
adrian-codecov, line 239: Could this be self.plan_data? 

I'm seeing a lot of `Plan.objects.get(name=self.plan_name)`, so this comment applies to all the instances that behave the same way

1) Could we use plan_data instead? Not sure how the account part of this works, but I'm not a fan of doing many DB queries in different properties when you could have made one. If we can't use plan_data, I'd make another field called "plan_db_representation" or something (please something shorter 😂), make that a cache property, and do that instead (this is effectively plan_data but idk how the accounts play a role).
2) Prefetch other relationships you have (tier below). When you do plan.tier, if you haven't prefetched that, you basically do another DB call to tier cause it's a fk, so prefetching here means "ask for it when you do the initial fetch cause it will be 1 query"

tests/unit/django_apps/codecov_auth/test_codecov_auth_models.py
```diff
@@ -396,6 +398,12 @@ def test_fields_that_account_overrides(self):
             ENTERPRISE_CLOUD_USER_PLAN_REPRESENTATIONS[self.owner.account.plan]
         )
         account_pretty_plan.update({"quantity": 0})
+        print(
```
adrian-codecov, line 401: 🖨️

tests/unit/django_apps/codecov_auth/test_codecov_auth_models.py
```diff
@@ -41,11 +41,13 @@
     PlanName,
 )
 from shared.utils.test_utils import mock_config_helper
+from tests.helper import mock_all_plans_and_tiers
 
 
 class TestOwnerModel(TransactionTestCase):
     def setUp(self):
         self.owner = OwnerFactory(username="codecov_name", email="name@codecov.io")
+        mock_all_plans_and_tiers()
```
adrian-codecov, line 50: Remind me, does setup run once per test or per class? If so, I'd try to put it in a method that does it once, no need to recreate it every time

tests/unit/upload/test_utils.py
```diff
@@ -18,9 +18,13 @@
     insert_coverage_measurement,
     query_monthly_coverage_measurements,
 )
+from tests.helper import mock_all_plans_and_tiers
 
 
 class CoverageMeasurement(TestCase):
+    def setUp(self):
+        mock_all_plans_and_tiers()
```
adrian-codecov, line 26: Why would we need that here?

shared/plan/constants.py
adrian-codecov: So why is this a circular dependency? I suppose cause you'd import Plan, and plan imports from constants.

So, this being the only thing that imports from Plan, could this belong in plan instead? You'd have to make an instance of Plan to use this wherever you do though

tests/unit/django_apps/codecov_auth/test_codecov_auth_models.py
ajay-sentry: That's a great point, apparently there's a setUpClass() function we might be able to use here. Not only for this setup but probably a lot of other tests as well

tests/unit/upload/test_utils.py
ajay-sentry: bc of this: https://github.com/codecov/shared/blob/b186b3c89fe16a4f9512cf160f39cfd262ba2eb8/shared/upload/utils.py#L35

admittedly could probably be more precise with the mock here

shared/plan/service.py
ajay-sentry: The full plan list we're porting over can be found here: https://www.notion.so/sentry/Plan-and-Tier-Table-Mapping-1758b10e4b5d801aa2d7e30198bcfc68

It's basically all the plans that were in USER_PLAN_REPRESENTATION, and seemingly the check should hold true even if the plans are no longer active in the future too, which is why I didn't add the is_active=True filter

shared/plan/constants.py
ajay-sentry: yeah exactly, I kinda wanted to avoid having to bind this function to a class when it's really just a "transformer" from our plan object to the plan object that the planService and subsequently GQL expects

shared/plan/service.py
ajay-sentry: oh yeah nice catch, saves a query too

shared/plan/service.py
ajay-sentry: Yeah those are great points, I was rushing through on the original implementation just to get us started on fixing tests but going back we can totally use plan_data instead, I think the account stuff will be needed as well so the stuff previously was wrong anyway.

2 birds with one stone: optimizations and bug fixing 😂 

I also fixed all the instances of plan.tier to now prefetch tier if we are gonna need it too

shared/plan/constants.py
```diff
@@ -189,7 +208,7 @@ def convert_to_DTO(self) -> dict:
             "Unlimited private repositories",
             "Priority Support",
         ],
-        tier_name=TierName.PRO.value,
+        tier_name=TierName.SENTRY.value,
```
ajay-sentry, line 211: note: These consts are all disappearing soon and aren't / shouldn't be a SOT for anything anymore

shared/plan/service.py
ajay-sentry: Added to all relevant touchpoints! Good call

tests/helper.py
```diff
@@ -46,3 +54,166 @@ def _sessions(sessions):
         "totals": report.get("totals", {}),
         "chunks": chunks,
     }
+
+
+def mock_all_plans_and_tiers():
```
ajay-sentry, line 59: Note: Eventually we want to migrate these off of using any enums at all; they really should be agnostic to whatever we have in production

RulaKhaled: Yep we could do that as part of the clean up bonus ticket

shared/plan/service.py
ajay-sentry: How can I confirm if its cached or not? Is there a way to check locally / on stage if we do a DB hit vs. use the cache with the new property? In any case I was able to update using that property in [7804571](https://github.com/codecov/shared/pull/479/commits/7804571e81d9195cab11307e02972cfdb5fe3eea)

shared/plan/service.py
ajay-sentry: Looked into this one a bit, @cached_property is unavailable to use because we have an additional input parameter (owner), and while LRU_cache does work, if we don't end up evicting the value for an owner when other aspects of that owner change we could end up with some stale / incorrect values

shared/plan/constants.py
```diff
@@ -73,6 +73,26 @@ class TierName(enum.Enum):
     TEAM = "team"
     PRO = "pro"
     ENTERPRISE = "enterprise"
+    SENTRY = "sentry"
+    TRIAL = "trial"
+
+
+def convert_to_DTO(plan) -> dict:
```
ajay-sentry, line 80: @Swatinem this guy

shared/plan/constants.py
adrian-codecov: I see. I'll leave the implementation up to you, I think though that not typing a function because of circular imports feels a bit odd, but you're getting of the fn soon right? If so I wouldn't say it's a big deal

shared/plan/service.py
adrian-codecov: Setting the decorator should suffice, there's probably a way to check it programatically as well but can't think off the top of my head

tests/unit/django_apps/codecov_auth/test_codecov_auth_models.py
adrian-codecov: Do we still want this print here or typo?

shared/django_apps/codecov_auth/models.py
```diff
@@ -605,7 +605,8 @@ def pretty_plan(self):
         if self.account:
             return self.account.pretty_plan
 
-        plan_details = Plan.objects.get(name=self.plan)
+        print("@@@@@@@@@@@@@@@@@@@@", self.plan)
```
adrian-codecov: Another print 🖨️

adrian-codecov (approved):

nora-codecov (approved):

PR #480, opened 2025-01-16
Bundle Analysis: Add asset comparison and fetching PR files

JerrySentry:
https://github.com/codecov/engineering-team/issues/3216
https://github.com/codecov/engineering-team/issues/3203

Create a way AssetComparison object to better compare assets, the old version only compares the size changes whereas this new class supports loading more comparison related info. Also adds the ability to fetch list of modules related to the assets being compared including a way to filter out files.

This supporting changed will be used by the upcoming PRs for worker to enable BA PR comments to show large file changes for changed assets.

### Legal Boilerplate


shared/bundle_analysis/parsers/v3.py
```diff
@@ -160,6 +161,16 @@ def parse(self, path: str) -> Tuple[int, str]:
                 # but first we need to find the Asset by the hashed file name
                 dynamic_imports_list = self._parse_dynamic_imports()
                 if dynamic_imports_list:
```
JerrySentry, line 163: Delete the old entries in the table before inserting the new data.

suejung-sentry (approved):
cool stuff!

PR #481, opened 2025-01-16
Migration: Add Stripe_id to Plan Model

ajay-sentry:
This PR adds a new column to the plan table for us to maintain a reference back to stripe (which is needed for paid plans)

Screenshot 2025-01-16 at 2 26 05 PM

### Legal Boilerplate


shared/django_apps/codecov_auth/migrations/0064_plan_stripe_id.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.2.16 on 2025-01-16 22:18
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
```
adrian-codecov, line 7: Here's where you'd add the output of sqlmigrate if you want 👍

adrian-codecov (approved):

shared/django_apps/codecov_auth/migrations/0064_plan_stripe_id.py
ajay-sentry: oooooo

PR #483, opened 2025-01-17
add storage_path index on ReportSession

joseph-sentry:


Swatinem (approved):
nice, this should also help with my deletion code, which uses a `IS NULL` query on this field to quickly delete all rows without one.

PR #486, opened 2025-01-24
dev: Add Ruff print rules to shared (T20)

ajay-sentry:
This PR aims to add the ruff print rules to worker, and either removes or ignores each print statement (I just used my best judgement)

Ruff rules can be found here: https://docs.astral.sh/ruff/rules/#flake8-print-t20

Closes https://github.com/codecov/engineering-team/issues/3277

### Legal Boilerplate

### DIFF
--- a/tests/unit/reports/test_carryforward.py
+++ b/tests/unit/reports/test_carryforward.py
@@ -1,5 +1,4 @@
 import dataclasses
-import pprint
 from json import loads
 
 import pytest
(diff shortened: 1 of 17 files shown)


trent-codecov (approved):

tests/unit/reports/test_carryforward.py
spalmurray-codecov, line 1: pprint, for when print isn't enough p

spalmurray-codecov (approved):

tests/unit/reports/test_carryforward.py
ajay-sentry: TIL that even existing ngl

PR #487, opened 2025-01-24
feat: Add Amplitude event publisher

spalmurray-codecov:
This PR adds a new `EventPublisher` abstraction and an implementation of it in `AmplitudeEventPublisher`. `EventPublisher` is meant to be extremely generic and could be reused for any types of events we may want to publish in the future (`TaskService` could be converted to an `EventPublisher`, for example).

The Amplitude implementation defines some starting events and their payloads. These will be used in a future API PR, once we have the API key set as an environment variable.

Closes https://github.com/codecov/engineering-team/issues/2893


shared/events/base.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC, abstractmethod
+
+
+class EventPublisherPropertyException(Exception):
```
ajay-sentry, line 4: MissingEventPropertyException

ajay-sentry (approved):
looks great! Nice work

PR #494, opened 2025-01-30
feat: Add new users-developer plan

RulaKhaled:
As part of milestone 4, we are introducing the new userrs-developer plan, which is basically a free version of a team plan. 

- replace basic plan occurrences with developer plan 
- add users-developer to owner's plan choices 
- default plan to developer
-  adjust plans.  

### Legal Boilerplate


shared/plan/constants.py
```diff
@@ -2,10 +2,13 @@
 from dataclasses import dataclass
 from typing import List, Optional
 
+DEFAULT_FREE_PLAN = "users-developer"
+
 
 class MonthlyUploadLimits(enum.Enum):
     CODECOV_BASIC_PLAN = 250
     CODECOV_TEAM_PLAN = 2500
+    CODECOV_DEVELOPER_PLAN = 250
```
ajay-sentry, line 11: Are we using this reference anywhere? We could probably rename "BASIC" to "FREE" and use that enum, or just use that enum instead of creating a new one (having duplicate keys tied to the same value is also a code smell)

shared/plan/constants.py
```diff
@@ -21,6 +24,7 @@ class PlanMarketingName(enum.Enum):
     BASIC = "Developer"
     TRIAL = "Developer"
     TEAM = "Team"
+    DEVELOPER = "Developer"
```
ajay-sentry, line 27: Same thought here, we could use Basic or remove if its not in use anywhere

shared/plan/constants.py
```diff
@@ -311,9 +316,25 @@ def convert_to_DTO(self) -> dict:
     monthly_uploads_limit=None,
 )
 
+DEVELOPER_PLAN = PlanData(
```
ajay-sentry, line 319: We shouldn't need these anymore right?

shared/plan/constants.py
```diff
@@ -311,9 +316,25 @@ def convert_to_DTO(self) -> dict:
...
+    ],
+    tier_name=TierName.TEAM.value,
+    trial_days=None,
+    monthly_uploads_limit=250,
+)
+
 FREE_PLAN_REPRESENTATIONS = {
     PlanName.FREE_PLAN_NAME.value: FREE_PLAN,
     PlanName.BASIC_PLAN_NAME.value: BASIC_PLAN,
+    DEFAULT_FREE_PLAN: DEVELOPER_PLAN,
```
ajay-sentry, line 337: We shouldn't need this anymore right?

shared/plan/constants.py
```diff
@@ -2,10 +2,13 @@
 from dataclasses import dataclass
 from typing import List, Optional
 
+DEFAULT_FREE_PLAN = "users-developer"
```
ajay-sentry, line 5: Can we put this next to the plan names so it's easier to see?

tests/helper.py
```diff
@@ -217,3 +218,18 @@ def mock_all_plans_and_tiers():
             "Priority Support",
         ],
     )
+
+    PlanFactory(
```
ajay-sentry, line 222: We could probably also get rid of the Basic_plan in this plan factory as well I hope?

tests/unit/plan/test_plan.py
```diff
@@ -243,31 +244,29 @@ def test_plan_service_start_trial_manually_already_on_paid_plan(self):
...
-            plan=PlanName.BASIC_PLAN_NAME.value,
+            plan=DEFAULT_FREE_PLAN,
             trial_start_date=trial_start_date,
             trial_end_date=trial_end_date,
         )
         plan_service = PlanService(current_org=current_org)
 
-        basic_plan = FREE_PLAN_REPRESENTATIONS[PlanName.BASIC_PLAN_NAME.value]
+        developer_plan = FREE_PLAN_REPRESENTATIONS[DEFAULT_FREE_PLAN]
         assert plan_service.current_org == current_org
```
ajay-sentry, line 258: A lot of these assertions probably aren't needed, we can probably just assert the plan name and call it a day

tests/unit/plan/test_plan.py
```diff
@@ -549,12 +547,12 @@ def test_available_plans_for_sentry_plan_non_trial(self, is_sentry_user):
 @freeze_time("2023-06-19")
 class AvailablePlansExpiredTrialLessThanTenUsers(TestCase):
     """
-    - users-basic, has trialed, less than 10 users -> users-pr-inappm/y, users-basic, users-teamm/y
-    - users-teamm/y, has trialed, less than 10 users -> users-pr-inappm/y, users-basic, users-teamm/y
-    - users-pr-inappm/y, has trialed, less than 10 users -> users-pr-inappm/y, users-basic, users-teamm/y
-    - sentry customer, users-basic, has trialed, less than 10 users -> users-pr-inappm/y, users-sentrym/y, users-basic, users-teamm/y
-    - sentry customer, users-teamm/y, has trialed, less than 10 users -> users-pr-inappm/y, users-sentrym/y, users-basic, users-teamm/y
-    - sentry customer, users-sentrym/y, has trialed, less than 10 users -> users-pr-inappm/y, users-sentrym/y, users-basic, users-teamm/y
+    - { DEFAULT_FREE_PLAN}, has trialed, less than 10 users -> users-pr-inappm/y,  DEFAULT_FREE_PLAN, users-teamm/y
```
ajay-sentry, line 550: While I understand these were probably all find and replaces, can we get rid of the extra spaces before DEFAULT_

tests/unit/plan/test_plan.py
```diff
@@ -723,9 +718,9 @@ def test_available_plans_for_sentry_plan_expired_trial_less_than_10_users(
 @freeze_time("2023-06-19")
 class AvailablePlansExpiredTrialMoreThanTenActivatedUsers(TestCase):
     """
-    - users-pr-inappm/y, has trialed, more than 10 activated users -> users-pr-inappm/y, users-basic
-    - sentry customer, users-basic, has trialed, more than 10 activated users -> users-pr-inappm/y, users-sentrym/y, users-basic
-    - sentry customer, users-sentrym/y, has trialed, more than 10 activated users -> users-pr-inappm/y, users-sentrym/y, users-basic
+    - users-pr-inappm/y, has trialed, more than 10 activated users -> users-pr-inappm/y,  DEFAULT_FREE_PLAN
```
ajay-sentry, line 721: Same comment here

tests/unit/plan/test_plan.py
```diff
@@ -918,11 +912,11 @@ def test_trial_not_started(self):
 class AvailablePlansOngoingTrial(TestCase):
     """
     Non Sentry User is trialing
-        when <=10 activated seats -> users-pr-inappm/y, users-basic, users-teamm/y
-        when > 10 activated seats -> users-pr-inappm/y, users-basic
+        when <=10 activated seats -> users-pr-inappm/y,  DEFAULT_FREE_PLAN, users-teamm/y
```
ajay-sentry, line 915: Same comment here

shared/billing/__init__.py
```diff
@@ -23,6 +23,7 @@ class BillingPlan(Enum):
     users_yearly = PlanName.CODECOV_PRO_YEARLY_LEGACY.value
     enterprise_cloud_monthly = PlanName.ENTERPRISE_CLOUD_MONTHLY.value
     enterprise_cloud_yearly = PlanName.ENTERPRISE_CLOUD_YEARLY.value
+    users_developer = PlanName.USERS_DEVELOPER.value
```
ajay-sentry, line 26: If this is deprecated do we need to add this here?

shared/plan/constants.py
ajay-sentry: thoughts on renaming this to DEFAULT_PLAN_NAME as well? Since there could be a world where this plan isn't free, and this is technically just the name of the plan and not the full object

shared/plan/constants.py
RulaKhaled: can do

shared/plan/constants.py
RulaKhaled: I feel like the name "Free" makes more sense here. It'll be easier to look for "Free" references if someone wants to update it. "Default plan name" could technically reference any other plan

RulaKhaled: I feel like the name "Free" makes more sense here. It'll be easier to look for "Free" references if someone wants to update it. "Default plan name" could potentially reference any other plan

shared/plan/constants.py
RulaKhaled: unfortunately we do because API is still using `FREE_PLAN_REPRESENTATIONS` once we clean up API we can go back and remove all plans representations.

shared/plan/constants.py
RulaKhaled: same thing here

tests/helper.py
RulaKhaled: yes!

tests/unit/plan/test_plan.py
RulaKhaled: i'll remove it so it's easier to remove duplications

shared/billing/__init__.py
RulaKhaled: yah it's deprecated just removed it

tests/unit/plan/test_plan.py
RulaKhaled: yah we're doing that in a lot of places, good thing to consider refactoring in a separate PR

shared/plan/constants.py
ajay-sentry: same response, we can double confirm with @adrian-codecov but that's what he told me originally

shared/plan/constants.py
ajay-sentry: What about the concept of "name" vs. "plan"? Would you say it's still clear that this is just a reference to the plan name and not the plan object?

ajay-sentry (approved):

PR #495, opened 2025-01-31
Add Pro plan via Migration

ajay-sentry:
This PR seeds the existing pro plan to the DB so we can use it for DEC and self-hosted by default. It also updates all owners to that plan. The migration only runs the logic if the env is enterprise. 

In the process we're also creating a new variable which will set the default plan name for an owner to be the pro plan if they're enterprise so they can avoid having to create a bunch of other plans in their DB.

Screenshot 2025-02-05 at 1 59 06 PM
Screenshot 2025-02-05 at 2 00 13 PM

### Legal Boilerplate


spalmurray-codecov (approved):
Lgtm

trent-codecov (approved):

adrian-codecov (approved):

nora-codecov (approved):
that one thing about using .update() but that's not a showstopper

PR #496, opened 2025-01-31
fix: has_seats_left property

joseph-sentry:
use current_org.activated_user_count instead of plan_activated_users since activated_user_count does not include users with student=True in its calculation


adrian-codecov (approved):
Seems reasonable. Was there a bug that happened that led to this ticket?

joseph-sentry:
> Seems reasonable. Was there a bug that happened that led to this ticket?

yes, but no ticket

ajay-sentry (approved):

PR #500, opened 2025-02-05
feat: move timeseries app from api to shared

joseph-sentry:
I want to be able to access the timeseries models in worker using Django.


adrian-codecov (approved):
Lgtm, have you been following this guide? https://www.notion.so/sentry/Migration-Guide-74712fa52fad404babc02543db72278b

I'd also check for timeseries related DB settings that you might be needing to add to Shared so these models also work in Worker.

And, as part of moving this, could you also remove basically everything you moved from API? The guide I believe should have documentation on how to do so as well

Swatinem (approved):
nice! I wanted to add this to the owner / repo delete task as well, so this will help with that.

PR #502, opened 2025-02-07
Optimize `match/_any`

Swatinem:
This introduces a stateful `Matcher` class that is used to implement the `match/_any` fns.

This can be used to avoid compiling regex matchers repeatedly.

### DIFF
--- a/shared/utils/match.py
+++ b/shared/utils/match.py
@@ -1,37 +1,65 @@
 import re
+from typing import Sequence
 
 
-def match(patterns, string):
-    if patterns is None or string in patterns:
-        return True
+class Matcher:
+    def __init__(self, patterns: Sequence[str] | None):
+        self._patterns = set(patterns or [])
+        self._is_initialized = False
+        self._positives: list[re.Pattern] = []
+        self._negatives: list[re.Pattern] = []
 
-    patterns = set([_f for _f in patterns if _f])
-    negatives = [a for a in patterns if a.startswith(("^!", "!"))]
-    positives = patterns - set(negatives)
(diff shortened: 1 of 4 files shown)


shared/utils/match.py
```diff
@@ -1,37 +1,65 @@
...
+            # did not match any required paths
             return False
 
-    if positives:
-        for pattern in positives:
-            # match was found
-            if re.match(pattern, string):
-                return True
+        else:
+            # no positives: everyting else is ok
```
ajay-sentry, line 47: typo: everything

shared/utils/match.py
```diff
@@ -1,37 +1,65 @@
...
+            return True
+
+    def match_any(self, strings: Sequence[str] | None) -> bool:
+        if not strings:
+            return False
+        return any(self.match(s) for s in strings)
```
ajay-sentry, line 53: does any() return true for the first truthy value or iterate the whole list? Maybe another opportunity for minor optimization there

shared/utils/match.py
```diff
@@ -1,37 +1,65 @@
...
 
 
-def match(patterns, string):
-    if patterns is None or string in patterns:
-        return True
+class Matcher:
+    def __init__(self, patterns: Sequence[str] | None):
+        self._patterns = set(patterns or [])
+        self._is_initialized = False
+        self._positives: list[re.Pattern] = []
```
ajay-sentry, line 9: took me a sec to understand what positives and negatives were, maybe a simple inline comment here would help ease any initial confusion for someone who stumbles upon this later

ajay-sentry (approved):
looks good! Just a couple small things

PR #504, opened 2025-02-07
Report caching

trent-codecov:
- Add Redis-based ttl for reports caching
- Modify build_report_from_commit to use Redis cache alongside archive service
- Update docker-compose to support configurable MinIO port
- Improve error handling and logging for cache operations
- Extend unit tests to cover new caching behavior and edge cases

### DIFF
--- a/shared/reports/api_report_service.py
+++ b/shared/reports/api_report_service.py
@@ -1,9 +1,13 @@
 import logging
+import pickle
+import sys
 
 import sentry_sdk
 from django.utils.functional import cached_property
+from prometheus_client import Gauge
 
 from shared.api_archive.archive import ArchiveService
+from shared.config import get_config
 from shared.django_apps.core.models import Commit
 from shared.helpers.flag import Flag
 from shared.reports.readonly import ReadOnlyReport as SharedReadOnlyReport
@@ -11,6 +15,10 @@ from shared.reports.resources import Report
 from shared.storage.exceptions import FileNotInStorageError
 
 log = logging.getLogger(__name__)
+
+report_size_gauge = Gauge(
+    "codecov_report_size_bytes", "Size of the report in bytes", ["report_type"]
+)
 
 
 class ReportMixin:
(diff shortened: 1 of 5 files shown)


tests/unit/reports/test_api_report_service.py
```diff
@@ -0,0 +1,204 @@
...
+            and args[2] == "cache_timeout"
+        ):
+            return 3600
+
+    mock_get_config.side_effect = side_effect
+    return mock_get_config
+
+
+@pytest.mark.django_db
+class TestBuildReportFromCommit:
```
michelletran-codecov, line 88: nit: We probably don't need a test class for this file. You can annotate all tests/classes with `django_db` in this file by adding this to the top:

```
pytestmark = pytest.mark.django_db
```

shared/reports/api_report_service.py
```diff
@@ -54,17 +63,67 @@ def build_report_from_commit(commit: Commit, report_class=None):
...
+            # Record size metrics
+            report_size = sys.getsizeof(report)
+            report_size_gauge.labels(report_type=report.__class__.__name__).set(
+                report_size
+            )
+
+            serialized_report = pickle.dumps(report)
+
+            try:
+                archive_service.write_file(cache_key, serialized_report)
```
michelletran-codecov, line 109: Will there be issues with writing over an existing file?

shared/reports/api_report_service.py
```diff
@@ -54,17 +63,67 @@ def build_report_from_commit(commit: Commit, report_class=None):
...
+
+    if cache_enabled:
+        cache_key = f"reports/cache/{commit.commitid}/{report_class.__name__ if report_class else 'SerializableReport'}"
+
+        redis_key = f"report:{commit.repository_id}:{commit.commitid}"
+        cached_report = cache.get(redis_key)
+        if cached_report:
+            # if ttl is not expired pull from archive
+            try:
+                cached_data = archive_service.read_file(cache_key)
```
michelletran-codecov, line 82: According to Sentry, it looks like `archive_service.read_file` might be a slow operation as well:

Screenshot 2025-02-07 at 2 35 49 PM

shared/reports/api_report_service.py
```diff
@@ -54,17 +63,67 @@ def build_report_from_commit(commit: Commit, report_class=None):
...
+        )
+
+        if report and cache_enabled:
+            # Record size metrics
+            report_size = sys.getsizeof(report)
+            report_size_gauge.labels(report_type=report.__class__.__name__).set(
+                report_size
+            )
+
+            serialized_report = pickle.dumps(report)
```
michelletran-codecov, line 106: Can we move this outside this `if` statement? It would be useful to get metrics for report_size generally (outside of this particular optimization).

michelletran-codecov (commented):
This might get us a bit of savings (looks like maybe 2s at most). Since this is behind a config, we can try testing it out.

shared/reports/api_report_service.py
trent-codecov: Unsure honestly. I am relying on the underlying lib for this. Let me see if I can test this.

shared/reports/api_report_service.py
trent-codecov: agreed. I am hoping the slowness is around the chunks logic rather than the actual read file. The trace isn't clear to me whether its the actual file download or other logic. The file for the client that prompted this is only 8mb. While large for a chunks file thats not a large file in general. May need to revisit the perf of this underlying lib.

shared/reports/api_report_service.py
```diff
@@ -54,17 +63,63 @@ def build_report_from_commit(commit: Commit, report_class=None):
...
+
+            try:
+                archive_service.write_file(cache_key, serialized_report)
+                # Set TTL for cache in redis
+                cache.set(
+                    redis_key,
+                    "cached",
+                    timeout=get_config(
+                        "setup", "report_service", "cache_timeout", default=3600
+                    ),
```
Swatinem, line 112: I might be a bit late to the party here, as this is already merged.

I don’t think this cache here makes any sense, and might even make things slightly worse:
- Redis has a TTL, but storage does not. this means you end up in a situation where you never end up deleting any of these cache files.
- The report classes are implemented in such a way that they are parsing things on-demand as much as possible. Since you are caching right after the initial `build_report`, next time you load this from cache, you will still have to parse the actual report file contents on-demand. So the cache is not helping for that particular case.
- The report classes are probably a lot larger in-memory (and thus when pickling the actual python objects) than on disk. So the serialized cache is likely larger than what you are caching in the first place. Not a huge issue, as presumably `pickle` is a bit quicker at deserializing than parsing from the other formats.
- Caching this as *one file* will likely save a bit vs having to load things from *two* files, so that might end up as a slight improvement.

PR #506, opened 2025-02-10
Bundle Analysis: add a getter for cache config

JerrySentry:
### Legal Boilerplate


shared/django_apps/bundle_analysis/service/bundle_analysis.py
```diff
@@ -13,3 +13,10 @@ def create_if_not_exists(repo_id, name, is_caching=True) -> None:
         CacheConfig.objects.get_or_create(
             repo_id=repo_id, bundle_name=name, defaults={"is_caching": is_caching}
         )
+
+    @staticmethod
+    def get_cache_option(repo_id, name) -> bool:
```
adrian-codecov, line 18: Nit: type repo_id and name

PR #507, opened 2025-02-10
feat: Disable Amplitude when not on prod

spalmurray-codecov:
This PR disables Amplitude when not on prod and also adds a new event type `User Logged in`.

### DIFF
--- a/shared/events/amplitude/publisher.py
+++ b/shared/events/amplitude/publisher.py
@@ -4,7 +4,7 @@ from typing import Union
 from django.conf import settings
 
 from amplitude import Amplitude, BaseEvent, Config, EventOptions
-from shared.environment.environment import get_current_env
+from shared.environment.environment import get_current_env, Environment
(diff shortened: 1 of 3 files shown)


shared/events/amplitude/publisher.py
```diff
@@ -27,7 +28,12 @@ class AmplitudeEventPublisher(EventPublisher):
 
     client: Amplitude
 
-    def __init__(self):
+    def __init__(self, override_env=False):
+        if get_current_env() != "production" and not override_env:
```
ajay-sentry, line 32: so we're also looking to disable for enterprise?

Wondering if we should also use the Enum -> https://github.com/codecov/shared/blob/640ffad28b94026e09066170ec32d93ec3702f24/shared/environment/environment.py#L11

spalmurray-codecov: Yes no amplitude on enterprise. and yes we should use that updating now

PR #508, opened 2025-02-10
feat: create testruns model in timeseries app

joseph-sentry:
creates the Testrun and TestrunSummary models for TA. We start by creating the regular table and then make it a hypertable, then we create the continuous aggregates, then set the cagg policies, then finally create the Testrun and TestrunBranchSummary


nora-codecov (approved):
just one q - do you want to do `RiskyRunSQL`? I'm not sure if we do that for timeseries things.

shared/django_apps/timeseries/migrations/0016_auto_20250206_1513.py
```diff
@@ -0,0 +1,42 @@
...
+                with (timescaledb.continuous) as
+                select
+                    repo_id,
+                    branch,
+                    testsuite,
+                    classname,
+                    name,
+                    min(computed_name) as computed_name,
+                    COUNT(DISTINCT CASE WHEN outcome = 'failure' OR outcome = 'flaky_fail' THEN commit_sha ELSE NULL END) AS cwf,
+                    time_bucket(interval '1 days', timestamp) as timestamp_bin,
```
Swatinem, line 27:
```suggestion
                    time_bucket(interval '1 days', timestamp) as timestamp_bin,
                    
                    COUNT(DISTINCT CASE WHEN outcome = 'failure' OR outcome = 'flaky_fail' THEN commit_sha ELSE NULL END) AS failing_commits,
```

please spell out `failing_commits` instead of the cryptic `cwf`.

also IMO its a bit more readable to visually group / separate the "group by" columns and the aggregates.

shared/django_apps/timeseries/migrations/0019_auto_20250206_1657.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.16 on 2025-02-06 16:57
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("timeseries", "0018_auto_20250206_1657"),
```
Swatinem, line 8: can you give these migrations a readable name?
also, it is possible to merge some of those together, or do they have to be separate migrations?

shared/django_apps/timeseries/migrations/0019_auto_20250206_1657.py
```diff
@@ -0,0 +1,23 @@
...
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            """
+            select add_continuous_aggregate_policy(
+                'timeseries_testrun_summary_1day',
+                start_offset => '7 days',
+                end_offset => '1 days',
```
Swatinem, line 17: is there some documentation of what these mean?
I thought we want to aggregate things for 60 days. what does the `7 days` here mean in relation to that?

shared/django_apps/timeseries/models.py
```diff
@@ -184,3 +185,94 @@ def is_backfilled(self) -> bool:
...
+        indexes = [
+            models.Index(
+                fields=["repo_id", "test_id", "flags_hash"],
+            ),
+        ]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["repo_id", "test_id", "flags_hash"],
+                name="flags_hash_test_id_unique",
+            ),
```
Swatinem, line 226: I believe a unique constraint already implicitly creates an index that can be used for queries. which means this would be duplicated.

on the other hand, I don’t see any indices being defined on the materialized views. Do we not need those?

shared/django_apps/timeseries/models.py
```diff
@@ -184,3 +185,94 @@ def is_backfilled(self) -> bool:
         if not self.created_at:
             return False
         return datetime.now() > self.created_at + timedelta(hours=1)
+
+
+class Testrun(ExportModelOperationsMixin("timeseries.testrun"), models.Model):
+    timestamp = models.DateTimeField(null=False, primary_key=True)
+
+    test_id = models.BinaryField(null=False)
+    flags_hash = models.BinaryField(null=True)
```
Swatinem, line 194: is the `flags_hash` not part of the `test_id`?
we also store all the flags inline as the `flags` array (which IMO is a better idea than having a join table)

do we need the `flags_hash` at all in that case then?

Swatinem (approved):
lgtm.

it would be nice to give the migrations a readable name, as well as rename the aggregate field `failing_commits` to make it more obvious.

otherwise, I would maybe add another index on `repo_id, branch` for the main testruns table.
The experiments I did last week using clickhouse made it clear that doing pre-aggregation on the high-cardinality `branch` (lots of feature branches, but very few runs for those each) are not making a ton of sense, and its possible to use the raw data table to query for per-test, or aggregate per-feature-branch.
The pre-aggregation / materialized view make sense for the main branch and "across all branches" though.

shared/django_apps/timeseries/models.py
joseph-sentry: > is the flags_hash not part of the test_id?
not anymore, i guess we can remove it since we're storing the flags in the table and we aren't using it in any of the CAggs

shared/django_apps/timeseries/migrations/0019_auto_20250206_1657.py
joseph-sentry: there's [docs here](https://docs.timescale.com/use-timescale/latest/continuous-aggregates/refresh-policies/#change-the-refresh-policy), we do want to aggregate for 60 days, but we won't be refreshing the aggregated rows past 7 days, so if we somehow process a seven day old upload again, we won't be refreshing the continuous aggregate with that data. 

I think we can tune this policy as we see fit going forward, this is kinda just a placeholder, and I don't think the start_offset even needs to be that far back.

joseph-sentry:
>The experiments I did last week using clickhouse made it clear that doing pre-aggregation on the high-cardinality branch (lots of feature branches, but very few runs for those each) are not making a ton of sense, and its possible to use the raw data table to query for per-test, or aggregate per-feature-branch.
The pre-aggregation / materialized view make sense for the main branch and "across all branches" though.

right, i just verified this locally with timescale and i think you're right, it also reflects what you saw with your testing in Alloy for the dedicated tests view.

joseph-sentry:
>just one q - do you want to do RiskyRunSQL? I'm not sure if we do that for timeseries things.

i think it should be fine in this case because from what i understand we usually do risky migrations for operations that will lock a table for a long time, like creating an index on a large table like uploads or repos. Since in this case we're creating a new table then running operations on it while its empty I think it should be fine

shared/django_apps/timeseries/migrations/0015_testrun_model.py
```diff
@@ -0,0 +1,111 @@
...
+        migrations.RunSQL(
+            "ALTER TABLE timeseries_testrun DROP CONSTRAINT timeseries_testrun_pkey;",
+            reverse_sql="",
+        ),
+        migrations.RunSQL(
+            "SELECT create_hypertable('timeseries_testrun', 'timestamp');",
+            reverse_sql="",
+        ),
+        migrations.AddIndex(
+            model_name="measurement",
```
Swatinem, line 58: this seems unrelated? might be good to move to a different PR/migration.

Swatinem (approved):

PR #509, opened 2025-02-10
ref: Remove plan constants in shared, create properties on model instead of dataclass/DTO

ajay-sentry:
This PR aims to remove the last of the "hardcoded" plan stuff that we don't need

Instead of that we have a few properties defined on the Plan model itself, which, since we're using "select_related" to fetch the tiers, we should avoid having an N+1 query

Closes https://github.com/codecov/engineering-team/issues/3256

### Legal Boilerplate


nora-codecov (approved):
I love deleting things

shared/django_apps/codecov_auth/models.py
```diff
@@ -1045,6 +1045,30 @@ class Meta:
     def __str__(self):
         return self.name
 
+    @property
```
RulaKhaled, line 1048: should we align these with properties we have in plan service? https://github.com/codecov/shared/blob/main/shared/plan/service.py#L326

ajay-sentry: yep def, good call

RulaKhaled (approved):

PR #510, opened 2025-02-11
ref: Remove BillingPlan enum and update references

ajay-sentry:
This PR aims to remove the deprecated BillingPlan enum and it's associated references throughout. I double checked worker and API and didn't see either used directly there so this seemed safe to me.

One thing that I couldn't quite figure out (maybe a SQLalchemy / Django ORM thing that was messing up) was on the test_router.py file duplicating the mock_all_plans_and_tiers() call vs. having a single one at the top of the file. I tried the setupClass and different fixture stuff but couldn't get it to work, figured it's a nice to have and not a dealbreaker

Depends on https://github.com/codecov/codecov-api/pull/1150

Closes https://github.com/codecov/engineering-team/issues/3346

### Legal Boilerplate


shared/celery_router.py
```diff
@@ -48,8 +48,8 @@ def route_tasks_based_on_user_plan(task_name: str, user_plan: str):
     default_task_queue = (
         route(task_name) or dict(queue=BaseCeleryConfig.task_default_queue)
     )["queue"]
-    billing_plan = BillingPlan.from_str(user_plan)
-    if is_enterprise_cloud_plan(billing_plan):
```
nora-codecov, line 52: does `user_plan` come from PlanService? it has the logic that gets the "correct" plan if owner has GL or Account

ajay-sentry: It's not from plan service but rather these helpers depending on the task: https://github.com/codecov/codecov-api/blob/258b8cba496f2aeaf6a8c11ae05c8b6c2a085c8c/services/task/task_router.py#L13

which seems kinda messy but I didn't wanna unwind it more than I had to haha

ajay-sentry: Worker stuff -> https://github.com/codecov/worker/blob/391700edbe263d6a033bcf3ffeffe311c8267fd6/celery_task_router.py#L155-L156

nora-codecov (approved):

PR #511, opened 2025-02-13
fix: handle Minio client retries better

joseph-sentry:
we're running into [this issue](https://codecov.sentry.io/issues/6293115637/events/11ab665a53664b6ca25b923a37eac0f7/) on Sentry, and given the docs [here](https://cloud.google.com/storage/docs/retry-strategy#python) i want to try and change some of our HTTP retry settings

### DIFF
--- a/shared/storage/minio.py
+++ b/shared/storage/minio.py
@@ -7,6 +7,8 @@ import tempfile
 from io import BytesIO
 from typing import BinaryIO, overload
 
+import certifi
+import urllib3
 from minio import Minio
 from minio.credentials import (
     ChainedProvider,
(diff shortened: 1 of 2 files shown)


shared/storage/minio.py
```diff
@@ -80,6 +85,25 @@ def init_minio_client(
         if port is not None:
             host = "{}:{}".format(host, port)
 
+        timeout = timedelta(minutes=5).seconds
+        http_client = urllib3.PoolManager(
+            timeout=Timeout(connect=timeout, read=timeout),
```
Swatinem, line 90: 5 minutes for `connect` is unreasonable. even for `read`, this might be a bit too generous.
for native symbolicator, we assumed a "low end" read throughput of ~4M/s, meaning that it fetches 1G within ~4 minutes.
as we are talking about *internal* storage here, we can set the expectations towards throughput a lot higher, and thus the timeouts a bit lower.

Swatinem (approved):
lgtm, though we should make the timeouts a lot tighter

PR #512, opened 2025-02-18
Add various benchmarks for the `Report` classes

Swatinem:
This benchmarks the following scenarios:
- Shallow parsing, he various `Report`s try to parse as little as possible initially
- Iterating over all the files and their lines within a report, which will parse the whole report
- Processing the `totals`, which internally iterates over everything and sums up things
- Report filtering and iterating over everything in a filtered Report
- Serializing
- Merging
- Creating a carryforward, which is a different way of creating a *fully filtered* report (the other filtering does things on-demand)

I get the following output when running with `pytest --codspeed`:

```
                                        Benchmark Results                                          
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━┓
┃                                       Benchmark ┃   Time (best) ┃ Rel. StdDev ┃ Run time ┃ Iters ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━┩
│                     test_report_parsing[Report] │       8,158ns │        0.8% │    2.95s │ 5,041 │
│             test_report_parsing[ReadOnlyReport] │      10,028ns │        2.0% │    2.97s │ 4,544 │
│        test_report_parsing[Rust ReadOnlyReport] │      10,149ns │        2.3% │    3.01s │ 4,544 │
│             test_report_parsing[EditableReport] │     211,784ns │        1.9% │    2.97s │   994 │
│                 test_report_iterate_all[Report] │  46,870,458ns │        1.5% │    2.97s │    63 │
│         test_report_iterate_all[ReadOnlyReport] │  47,397,167ns │        0.9% │    2.97s │    62 │
│    test_report_iterate_all[Rust ReadOnlyReport] │  47,578,833ns │        1.3% │    2.97s │    62 │
│         test_report_iterate_all[EditableReport] │  45,093,959ns │        1.6% │    2.95s │    65 │
│              test_report_process_totals[Report] │ 104,394,875ns │        0.9% │    2.95s │    28 │
│      test_report_process_totals[ReadOnlyReport] │ 104,967,458ns │        1.5% │    2.99s │    28 │
│ test_report_process_totals[Rust ReadOnlyReport] │ 104,451,541ns │        1.1% │    2.97s │    28 │
│      test_report_process_totals[EditableReport] │ 102,963,667ns │        1.3% │    2.93s │    28 │
│                   test_report_filtering[Report] │ 224,757,166ns │        0.2% │    2.93s │    13 │
│           test_report_filtering[ReadOnlyReport] │ 235,387,583ns │        0.8% │    2.84s │    12 │
│      test_report_filtering[Rust ReadOnlyReport] │ 122,185,791ns │        6.6% │    3.05s │    24 │
│           test_report_filtering[EditableReport] │ 220,443,292ns │        0.7% │    2.88s │    13 │
│                   test_report_serialize[Report] │       2,295ns │        2.7% │    2.96s │ 9,443 │
│           test_report_serialize[EditableReport] │   2,905,792ns │        0.2% │    2.94s │   252 │
│                       test_report_merge[Report] │ 243,114,375ns │        0.4% │    2.95s │    12 │
│               test_report_merge[EditableReport] │ 253,787,500ns │        1.5% │    2.82s │    11 │
│                test_report_carryforward[Report] │   2,728,568ns │        1.6% │    2.98s │   268 │
│        test_report_carryforward[EditableReport] │   2,708,930ns │        1.7% │    2.96s │   268 │
└─────────────────────────────────────────────────┴───────────────┴─────────────┴──────────┴───────┘
```

Some interesting results:
- The `EditableReport` seems to do a bit more work at parse time, though the `merge` time is within 5% of `Report`
- Initially I thought I was doing something wrong related to the `should_load_rust_version`, but it looks like it only offers a speedup for the `filter` case.
- The `Report` serialize timing looks like a fluke. Not sure whats up with that, in particular compared to `EditableReport`.

Otherwise, the difference between the various report types looks very much like noise.

---

While looking a bit more into the various `Report` classes, here are some other observations:
- In *most* of the cases (except maybe `merge`), the various iterators only yield temporary copied of the fully-parsed `ReportFile` and `ReportLine` classes. Meaning that if you iterate over the report twice, you re-parse everything.
- The code to maintain cached versions of `totals` is quite weird in a bunch of places.
- In particular, now that we have parallel processing, we do not need to maintain uptodate `totals` for the *intermediate* reports at all. And when merging multiple reports, we are updating the `totals` on each merge step, even though refreshing those once at the end when serializing would be enough.
- ☝🏻 this might be a good idea for some reasonably low hanging fruit for optimization


tests/benchmarks/test_report.py
```diff
@@ -0,0 +1,141 @@
...
+
+@pytest.mark.parametrize("report_class, should_load_rust", READABLE_VARIANTS)
+def test_report_filtering(report_class, should_load_rust, mocker, benchmark):
+    raw_chunks, raw_report_json = init_mocks(mocker, should_load_rust)
+
+    report = do_parse(report_class, raw_report_json, raw_chunks)
+
+    def bench_fn():
+        filtered = report.filter(paths=[".*"], flags=["unit"])
+        filtered._process_totals()
```
giovanni-guidini, line 95: Are these operations necessary for the filtering, or the filter process is a lazy one and you're just forcing the filtered report being parsed?

Swatinem: yes, the second one. the `filter` method just creates a wrapper class based on the underlying report, with a different filter on top.
and getting the totals (or iterating the files) is one of the operations that is frequently performed on those reports.

for example, the whole slowness of component comparisons is based on getting the totals / diff totals for the report filtered for the specific component / flag

giovanni-guidini (approved):

PR #513, opened 2025-02-18
feat: Add required event properties for Amplitude upload event

spalmurray-codecov:
Adds the required event properties for the Amplitude 'Upload Sent' event that will be instrumented in API after this is merged.


shared/events/amplitude/publisher.py
```diff
@@ -118,7 +118,15 @@ def set_group(
         group_name: Union[str, list[str]],
         event_options: EventOptions,
     ):
+        log.info(
```
spalmurray-codecov, line 121: these logs make it easier to validate payloads without actually having amplitude configured locally.

RulaKhaled (approved):

PR #517, opened 2025-02-19
ci: use same coverage/test result filenames as other repos

matt-codecov:
useful for https://github.com/codecov/umbrella/pull/4

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,23 +45,23 @@ jobs:
       - name: upload using codecovcli
         if: ${{ !cancelled() && !github.event.pull_request.head.repo.fork && github.repository_owner == 'codecov' }}
         run: |
-          docker-compose exec shared codecovcli -v upload-process --flag shared-docker-uploader --token ${{ secrets.CODECOV_ORG_TOKEN }} --fail-on-error
-          docker-compose exec shared codecovcli -v do-upload --report-type "test_results" --flag shared-docker-uploader --token ${{ secrets.CODECOV_ORG_TOKEN }} --fail-on-error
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -45,23 +45,23 @@ jobs:
       - name: upload using codecovcli
         if: ${{ !cancelled() && !github.event.pull_request.head.repo.fork && github.repository_owner == 'codecov' }}
         run: |
-          docker-compose exec shared codecovcli -v upload-process --flag shared-docker-uploader --token ${{ secrets.CODECOV_ORG_TOKEN }} --fail-on-error
-          docker-compose exec shared codecovcli -v do-upload --report-type "test_results" --flag shared-docker-uploader --token ${{ secrets.CODECOV_ORG_TOKEN }} --fail-on-error
+          docker-compose exec shared codecovcli -v upload-process --flag shared-docker-uploader --file tests/unit.coverage.xml --token ${{ secrets.CODECOV_ORG_TOKEN }} --fail-on-error
```
Swatinem, line 48: Does `shared` not have the distinction between unit and integration tests, like the other repos do?

Swatinem (approved):

.github/workflows/ci.yml
matt-codecov: i think some tests call themselves integration tests but it's not worked into the repo/CI currently. we should change that, and other things, so that we can reuse the workflows worker/api get to use

currently i am disturbing as little as possible so get things working in umbrella, and then we can assess whether to change shared to follow that pattern or change all three projects to use a more monorepo-aware CI structure

PR #519, opened 2025-02-20
Clean up `delete_session` related code

Swatinem:
This removes the existing implementation on `Report`, as only the `EditableReport` is supposed to have them. It also removes the single `delete_session` in favor of the `delete_multiple_sessions` method.

The remaining method is optimized slightly.

### DIFF
--- a/tests/unit/reports/test_editable.py
+++ b/tests/unit/reports/test_editable.py
@@ -73,15 +73,15 @@ def test_merge_coverage():
 class TestEditableReportHelpers(object):
     def test_line_without_session(self):
         line = ReportLine.create(1, None, [LineSession(1, 0), LineSession(0, 1)])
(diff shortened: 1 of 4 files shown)


tests/unit/reports/test_editable.py
```diff
@@ -1674,8 +1674,7 @@ def test_delete_session(self, sample_report):
         assert res["report"] == expected_result["report"]
         assert res["totals"] == expected_result["totals"]
         assert res == expected_result
-        report.delete_session(0)
-        report.delete_session(2)
+        report.delete_multiple_sessions({0, 2})
```
adrian-codecov, line 1677: Nice, was going to ask if there was a test testing multiple sessions, had only seen a dict with 1, tyty

adrian-codecov (approved):

PR #520, opened 2025-02-20
Add benchmarks to CI workflow

Swatinem:


Swatinem:
Interesting.

The first run in valgrind / instrumentation mode was unbearably slow, ~17 minutes vs <2 min locally.
"Downgrading" to the walltime based measurements makes the CI run quite a bit faster, taking "only" 3 minutes, on par with the normal testsuite.

I see both runs in https://codspeed.io/codecov/shared/runs, but the second "walltime" run does not properly load 🤔

adrian-codecov (approved):

PR #522, opened 2025-02-20
feat: Add anonymous user details for amplitude

spalmurray-codecov:
Adds constants for using an anonymous user with Amplitude.

### DIFF
--- a/shared/events/amplitude/__init__.py
+++ b/shared/events/amplitude/__init__.py
@@ -1,3 +1,4 @@
 from shared.events.amplitude.publisher import AmplitudeEventPublisher
+from shared.events.amplitude.constants import UNKNOWN_USER_OWNERID
 
-__all__ = ["AmplitudeEventPublisher"]
+__all__ = ["AmplitudeEventPublisher", "UNKNOWN_USER_OWNERID"]
--- a/shared/events/amplitude/__init__.py
+++ b/shared/events/amplitude/__init__.py
@@ -1,4 +1,4 @@
-from shared.events.amplitude.publisher import AmplitudeEventPublisher
 from shared.events.amplitude.constants import UNKNOWN_USER_OWNERID
+from shared.events.amplitude.publisher import AmplitudeEventPublisher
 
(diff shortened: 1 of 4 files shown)


shared/events/amplitude/publisher.py
```diff
@@ -47,8 +49,15 @@ def __init__(self, override_env=False):
     def publish(
         self, event_type: AmplitudeEventType, event_properties: AmplitudeEventProperties
     ):
+        user_id = event_properties["user_ownerid"]
+
         # Handle special set_orgs event
         if event_type == "set_orgs":
+            if user_id == UNKNOWN_USER_OWNERID:
+                # We don't want to track these events for unknown users,
```
ajay-sentry, line 57: is there any value in adding a log here in the event it does happen?

shared/events/amplitude/__init__.py
```diff
@@ -1,3 +1,4 @@
+from shared.events.amplitude.constants import UNKNOWN_USER_OWNERID
 from shared.events.amplitude.publisher import AmplitudeEventPublisher
 
-__all__ = ["AmplitudeEventPublisher"]
+__all__ = ["AmplitudeEventPublisher", "UNKNOWN_USER_OWNERID"]
```
ajay-sentry, line 4: do we need to export from here if we're importing from the constants file anyway?

shared/events/amplitude/publisher.py
```diff
@@ -5,6 +5,7 @@
 
 from amplitude import Amplitude, BaseEvent, Config, EventOptions
 from shared.environment.environment import Environment, get_current_env
+from shared.events.amplitude.constants import UNKNOWN_USER_OWNERID
```
ajay-sentry, line 8: otherwise we can change this to drop .constants

ajay-sentry (approved):
lgtm, just some minor things

shared/events/amplitude/publisher.py
spalmurray-codecov: Don't think so honestly. It really shouldn't happen. If it does, I don't think we care.

ajay-sentry: Fair enough

PR #525, opened 2025-02-24
Add a benchmark for `apply_diff`

Swatinem:
I’m actually surprised that it is reasonably quick. But then I remembered that the runtime depends on the size of the diff, not on the size of the report, so things scale differently.

```
                                          Benchmark Results                                          
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━┓
┃                                       Benchmark ┃   Time (best) ┃ Rel. StdDev ┃ Run time ┃  Iters ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━┩
│            test_report_diff_calculation[Report] │         973ns │        4.4% │    2.98s │ 14,555 │
│    test_report_diff_calculation[FilteredReport] │       1,915ns │        1.9% │    2.92s │ 10,295 │
└─────────────────────────────────────────────────┴───────────────┴─────────────┴──────────┴────────┘
```


giovanni-guidini (approved):
The standard deviation for the regular report is somewhat high compared to the filtered report. I wonder why that is... if it's the same process repeated many times.

Swatinem:
You are right, yes. Not quite sure why that is. Maybe its just a fluke when I ran this locally, but its also possible that some internal caching is going on there.

PR #526, opened 2025-02-24
Remove unused `ReadOnlyReport.append`

Swatinem:
This method seemed to have been intended to warn about wrong usage of `ReadOnlyReport`.

I checked both in production logs, as well as scanning through the code, and it looks like we are not misusing the this report class in such a way. So it is safe to remove this warning.


giovanni-guidini (approved):
Would we benefit from throwing an exception here instead of hoping that we will keep not misusing the `ReadOnlyReport`?

EDIT ah no I see `ReadOnlyReport` is not a subclass of any other report, so removing it would cause an exception anyway

PR #527, opened 2025-02-24
Remove unused filtering code from `Report/File`

Swatinem:
The base `Report` and `ReportFile` classes had some code related to filtering which was never actually used, as filtering works using the `FilteredReport` class.

### DIFF
--- a/shared/reports/resources.py
+++ b/shared/reports/resources.py
@@ -54,7 +54,6 @@ class ReportFile(object):
         "name",
         "_details",
         "_lines",
-        "_line_modifier",
         "_ignore",
         "_totals",
     ]
@@ -64,7 +63,6 @@ class ReportFile(object):
         name,
         totals=None,
         lines=None,
-        line_modifier=None,
         ignore=None,
     ):
         """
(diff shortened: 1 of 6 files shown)


shared/reports/resources.py
```diff
@@ -93,18 +90,13 @@ def __init__(
             self._details = {}
             self._lines = []
 
-        # <line_modifier callable>
-        self._line_modifier = line_modifier
         self._ignore = _ignore_to_func(ignore) if ignore else None
 
-        if line_modifier:
-            self._totals = None  # need to reprocess these
+        # totals = <ReportTotals []>
```
giovanni-guidini, line 95: is this supposed to mean `totals: list[ReportTotals]` ?

It seems is more like `ReportTotals | list | None`, where `list` is the serialized version of `ReportTotals`

My point is the comment is a bit useless

shared/reports/resources.py
```diff
@@ -876,31 +806,18 @@ def _process_totals(self):
...
-        _path_filter = None or bool
-        _line_modifier = self._line_modifier
 
         def _iter_totals():
             for filename, data in self._files.items():
-                if not _path_filter(filename):
-                    continue
-                elif _line_modifier or data.file_totals is None:
+                if data.file_totals is None:
                     # need to rebuild the file because there are line filters
```
giovanni-guidini, line 813: is this case still possible if you removed all the filters?

giovanni-guidini (approved):

PR #528, opened 2025-02-24
Consolidate implementation of `calculate_diff`

Swatinem:
This cleans up the duplicated implementations of `Filtered/Report.calculate_diff` and moves it to a new well typed file.

---

I’m a bit surprised that this does not yield any speedup in the diff calculation benchmarks (compared to https://github.com/codecov/shared/pull/525).

I think the difference between the base and `FilteredReport` is based on the actual filtering and line-mapping code that happens, instead of the different and less optimized version that was previously there.

```
                                          Benchmark Results                                          
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━┓
┃                                       Benchmark ┃   Time (best) ┃ Rel. StdDev ┃ Run time ┃  Iters ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━┩
│            test_report_diff_calculation[Report] │         803ns │        0.9% │    2.92s │ 15,975 │
│    test_report_diff_calculation[FilteredReport] │       1,886ns │        1.1% │    2.92s │ 10,437 │
└─────────────────────────────────────────────────┴───────────────┴─────────────┴──────────┴────────┘
```

### DIFF
--- a/shared/reports/diff.py
+++ b/shared/reports/diff.py
@@ -1,0 +1,88 @@
+from typing import Generator, Protocol, TypedDict
+
+from shared.reports.totals import get_line_totals
+from shared.reports.types import ReportLine, ReportTotals
+from shared.utils.totals import sum_totals
+
+
+class DiffSegment(TypedDict):
+    lines: list[str]
+    header: tuple[int, int, int, int]
+
+
+class DiffFile(TypedDict):
+    type: str
+    segments: list[DiffSegment]
+
+
+class RawDiff(TypedDict):
+    files: dict[str, DiffFile]
+
+
+class CalculatedDiff(TypedDict):
+    general: ReportTotals
+    files: dict[str, ReportTotals]
+
+
+# TODO: it might make sense to move these abstract interfaces to a different place
+class AbstractReportFile(Protocol):
+    def get(self, line_no: int) -> ReportLine | None: ...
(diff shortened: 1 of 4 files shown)


giovanni-guidini (approved):
LGTM. Thanks for adding more types to this part of the code 🥳 

As you mentioned no speed improvement, but in the comparison the stddev for the `Report` class was greatly reduced... that might be a win? Having a more bounded code I guess. The speed report doesn't really report quantiles for us to know, though.

I wonder if using `@dataclass` instead of `TypedDict` could improve speed. But my guess is that it would be worse (cause you need to transform from dict into a class instance).

Swatinem:
Thanks for the suggestions. I improved the types and added doc comments.

On that note, I just hate how python doc comments come *after* the item they are documenting. This is so unintuitive compared to pretty much *every other language* that does it differently.

Bildschirmfoto 2025-02-26 um 10 37 12

PR #531, opened 2025-02-24
feat: Make Amplitude publish fail gracefully

spalmurray-codecov:
This PR adds an `_unsafe_publish` method to the Amplitude event publisher, converting the existing `publish` method to be safe by default (i.e., it will always fail gracefully).

Also adds prometheus metrics to track success and failures of amplitude publish calls.

### DIFF
--- a/shared/events/amplitude/publisher.py
+++ b/shared/events/amplitude/publisher.py
@@ -49,6 +49,15 @@ class AmplitudeEventPublisher(EventPublisher):
     def publish(
         self, event_type: AmplitudeEventType, event_properties: AmplitudeEventProperties
     ):
+        try:
+            self.unsafe_publish(event_type, event_properties)
+        except Exception as e:
+            log.error("Failed to publish Amplitude event", extra=dict(error=str(e)))
+
+    def unsafe_publish(
+
+        self, event_type: AmplitudeEventType, event_properties: AmplitudeEventProperties
+    ):
         user_id = event_properties["user_ownerid"]
 
         # Handle special set_orgs event
--- a/shared/events/amplitude/publisher.py
+++ b/shared/events/amplitude/publisher.py
@@ -6,7 +6,10 @@ from django.conf import settings
 from amplitude import Amplitude, BaseEvent, Config, EventOptions
 from shared.environment.environment import Environment, get_current_env
 from shared.events.amplitude import UNKNOWN_USER_OWNERID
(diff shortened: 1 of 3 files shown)


shared/events/amplitude/publisher.py
```diff
@@ -48,6 +53,24 @@ def __init__(self, override_env=False):
...
+            },
+        )
+        try:
+            self._unsafe_publish(event_type, event_properties)
+        except Exception as e:
+            inc_counter(
+                AMPLITUDE_PUBLISH_FAILURE_COUNTER,
+                labels={"event_type": event_type, "error": e.__class__.__name__},
+            )
+            log.error("Failed to publish Amplitude event", extra=dict(error=str(e)))
```
spalmurray-codecov, line 70: can possibly remove this. Lmk thoughts

ajay-sentry: I like having the error obj here, and in fact I'd add the error_name too for easier filtering like this: https://github.com/codecov/codecov-api/blob/596618acd3b14210d83a98dd0917b16f6ce0be92/core/commands/commit/interactors/get_file_content.py#L31

ajay-sentry (approved):

shared/events/amplitude/publisher.py
spalmurray-codecov: done. Added the event_type also

shared/events/amplitude/publisher.py
```diff
@@ -48,6 +53,31 @@ def __init__(self, override_env=False):
...
+            )
+            log.error(
+                "Failed to publish Amplitude event",
+                extra=dict(
+                    event_type=event_type,
+                    error_name=e.__class__.__name__,
+                    error_message=str(e),
+                ),
+            )
+
```
nora-codecov, line 78: `except Exception` is broad - every type of raised error will qualify. Can you make it more specific?
If not, it looks like you're swallowing the error (nothing is raised) - is this how you want it to work?

nora-codecov (approved):
left some thoughts, other than that lgtm!

shared/events/amplitude/publisher.py
spalmurray-codecov: This is the intention yes. The only code below this is the little logic in this file and the underlying amplitude library. We are adding the error details to the log and the counter, so in theory we can catch unexpected errors getting swallowed here. 

The end goal of this PR is to make certain calling `publish` will never error out so we can put it wherever we want with 0 risk of interrupting critical flows.

nora-codecov: I see, that makes sense. awesome!

PR #533, opened 2025-02-24
fix: make plan non-nullable and drop plan type

ajay-sentry:
This PR is an extension of https://github.com/codecov/worker/pull/1103, closing the loop on plan type thats nullable as well as removing the default value for the plan which was a plan name we don't want to support anymore. 

This will hopefully prevent others from falling into the same pitfall we did 

### Legal Boilerplate


nora-codecov (approved):

adrian-codecov (commented):
Pending confirmation if these should be risky changes and what is the effect from making a plan non-nullable

PR #539, opened 2025-02-26
Set `complexity/_total` to `None` in diff calculation

Swatinem:
This was a recent regression from consolidating the diff calculation. Downstream code expects/asserts the `complexity/_total` to be `None` when no lines have been touched by the diff.

### DIFF
--- a/shared/reports/diff.py
+++ b/shared/reports/diff.py
@@ -1,3 +1,4 @@
+import dataclasses
 from typing import Generator, Literal, Protocol, TypedDict
 
 from shared.reports.totals import get_line_totals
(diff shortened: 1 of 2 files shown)


shared/reports/diff.py
```diff
@@ -97,4 +98,9 @@ def calculate_report_diff(report: AbstractReport, diff: RawDiff) -> CalculatedDi
 
     totals = sum_totals(list_of_file_totals)
 
+    if totals.lines == 0:
+        totals = dataclasses.replace(
```
joseph-sentry, line 102: if there's a reason we're using dataclasses.replace instead of in place modifying the attrs of totals can you mention it in the PR description

joseph-sentry (approved):
lgtm, just a request for clarification for the PR description

shared/reports/diff.py
Swatinem: I haven’t really thought about this. I just restored the code as it existed before my refactor, so there is no functional change.

PR #541, opened 2025-02-26
refactor: use dict {} syntax for typing

joseph-sentry:
getting a type error with the TypedDict when using the dict syntax to create the ProviderPull in github.py and using the {} syntax fixes this


adrian-codecov (approved):
Wait really? What makes the non-dict approach work w/ typedict?

Swatinem (approved):
I think the `dict` fn obscures the resulting type, whereas the literal makes it very obvious what it is.

PR #542, opened 2025-02-26
feat: remove usage of use_minio feature flag

joseph-sentry:
this rollout is done and we are 100% enabled on prod so we are safe to remove this and simplify the logic

starting off by removing all references to use_minio in shared


shared/storage/__init__.py
```diff
@@ -18,10 +18,7 @@ def get_appropriate_storage_service(
     if force_minio:
         chosen_storage = "minio"
 
-    if repoid:
-        if USE_MINIO.check_value(repoid, default=False):
-            chosen_storage = "minio"
-
+    if repoid and chosen_storage == "minio":
```
Swatinem, line 21: As we are defaulting/hardcoding `minio` now, you should rather remove the `chosen_storage` check here. Or rather initializing it as `minio`, independent of the `get_config` and the `force_minio` that I recently introduced.

Swatinem (changes requested):
I think this is not yet correct. But lets goooo! 🎉

shared/storage/__init__.py
joseph-sentry: are you suggesting doing
```
chosen_storage = "minio"
if repoid:
  if USE_NEW_MINIO...
 return _storage_service_cache...
```

Swatinem: yes, pretty much

PR #545, opened 2025-02-26
fix ci: add fetch-depth: 2 to checkout step to fix codecov

matt-codecov:
github puts a merge commit at the top of every PR. the CLI is supposed to detect that and use the actual PR head as the SHA instead of the merge commit, but because we were missing this fetch-depth argument, the local checkout only had the merge commit in it.

### Legal Boilerplate

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
       - name: Install docker compose
         run: |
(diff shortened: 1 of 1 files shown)


spalmurray-codecov (approved):

Swatinem:
We recently chatted around the concept of "adapt our product so it works with the customers setup", vs "the customer has to update their CI/codebase to work with our product".

The PR description is good, but I would like to better understand why we can’t just fix this case in the CLI, or server side for that matter?
I don’t think we can ever explain all customers that they would have to add this to their CI. So is everyone broken in some way now? Or only folks using the github merge queue?

PR #546, opened 2025-02-26
fix: improve new minio implementation

joseph-sentry:
The team mentioned wanting to ship the new read implementation first
then the new write implementation so this commit enables that.

i'm folding the NewMinioStorageService functionality back into the
MinioStorageService and creating 2 new feature flags one for read and
one for write.

Based on the values of those feature flags the MinioStorageService
will choose an implementation of read/write.

I also removed the old USE_NEW_MINIO feature flag, and modified the
get_appropriate_storage_service function to make it so it no longer
caches the MinioStorageService and instead the MinioStorageService is
caching its internal minio_client since that is the expensive part
of creating a new MinioStorageService instance.

I also moved some code around


shared/storage/__init__.py
```diff
@@ -14,43 +10,20 @@ def get_appropriate_storage_service(
...
-
-    if chosen_storage not in _storage_service_cache:
-        _storage_service_cache[chosen_storage] = (
-            _get_appropriate_storage_service_given_storage(chosen_storage)
+    return get_minio_storage_service(repoid)
+
+
+def get_minio_storage_service(
+    repo_id: int | None,
+) -> MinioStorageService:
```
Swatinem, line 18: might as well just inline this code

shared/storage/minio.py
```diff
@@ -1,133 +1,158 @@
...
+    def __init__(
+        self,
+        minio_config,
+        new_read: bool = False,
+        new_write: bool = False,
+    ):
+        self.minio_config = minio_config
+        self.new_read = new_read
+        self.new_write = new_write
+        self.zstd_default = zstd_decoded_by_default()
```
Swatinem, line 146: this should probably be moved to a global, as checking with `importlib` every time this is created is probably a very bad idea.

shared/storage/minio.py
```diff
@@ -1,133 +1,158 @@
...
+
+        log.debug("Connecting to minio with config %s", self.minio_config)
+
+        self.minio_client = get_cached_minio_client(**self.minio_config)
+
+        log.debug("Done setting up minio client")
+
+    def client(self):
+        return self.minio_client if self.minio_client else None
```
Swatinem, line 155: the `minio_client` is always defined, so this will never return `None`?
Why do we need this method at all in that case?

shared/storage/read.py
```diff
@@ -0,0 +1,82 @@
...
+    except MinioException:
+        raise
+
+
+def new_minio_read(
+    minio_client: Minio,
+    bucket_name: str,
+    path: str,
+    file_obj: IO[bytes] | None = None,
+    zstd_default: bool = False,
```
Swatinem, line 44: you are never passing this parameter

Swatinem (approved):

PR #547, opened 2025-02-27
Update the python version, and `uv.lock`

Swatinem:
Bumps the `.python-version` to `3.13`, and adds a required `uv` version as well, along with a `uv sync` that trims a bunch of python 3.12 wheel from the lockfile.


giovanni-guidini (approved):
This will likely break the api build process that uses a python 3.12 image (see [here](https://github.com/codecov/codecov-api/blob/main/docker/Dockerfile.requirements))

Just something to keep in mind.

ps: I might be wrong, but the default arg for python image _is_ `ARG PYTHON_IMAGE=python:3.12-slim-bookworm`

Swatinem:
Oh, that is a good point. I didn’t know that API was still on python 3.12, I thought we had updated to 3.13 all across the board.

On another note, this update broke the benchmark run (which does not work properly yet anyway), and it looks like I missed the following from https://docs.codspeed.io/benchmarks/python#usage-with-uv:

> Using actions/setup-python to install python and not uv install is critical for tracing to work properly.

PR #548, opened 2025-02-27
Merge `EditableReportFile` into `ReportFile`

Swatinem:
The main purpose of the `EditableReportFile` was to offer methods to delete labels/sessions, as well as to properly maintain uptodate versions of `present_sessions` and `totals` when doing mutations on the file.

All the methods were now moved to `ReportFile`, which has also gained new functionality related to making sure that caches are cleared when necessary, and re-computed on-demand, so that they are always correct.

---

I plan to do a similar thing for `EditableReport` as well in a followup, as it has a similar purpose in relation to maintaining uptodate `totals`.

This is part of https://github.com/codecov/engineering-team/issues/3381

### DIFF
--- a/shared/reports/resources.py
+++ b/shared/reports/resources.py
@@ -56,19 +56,11 @@ END_OF_HEADER = "\n<<<<< end_of_header >>>>>\n"
 
 
 class ReportFile(object):
-    __slots__ = [
-        "name",
-        "_details",
-        "_lines",
-        "_ignore",
-        "_totals",
-    ]
-
     def __init__(
         self,
-        name,
-        totals=None,
-        lines=None,
+        name: str,
+        totals: ReportTotals | list | None = None,
+        lines: list[None | str | ReportLine] | str | None = None,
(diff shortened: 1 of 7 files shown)


shared/reports/resources.py
```diff
@@ -424,6 +454,38 @@ def line_without_labels(
             sessions=new_sessions,
         )
 
+    def delete_labels(
```
adrian-codecov, line 457: Is it worth documenting somehow that these functionalities used to belong to an EditableReport and that should be used only in particular contexts? I guess we're not making that distinction any more since we're merging Report related classes, so I guess it's fair game to use any property/methods since they're moved to the ReportFile

adrian-codecov (approved):

shared/reports/resources.py
Swatinem: Yes, it should be fair game to use any method now.
In reality, all the mutating methods were already part of the previous interface anyway.

PR #549, opened 2025-02-27
Merge `EditableReport` into `Report`

Swatinem:
The primary usecase for `EditableReport` was to always keep a list of "partially parsed" `ReportFile`s around, as well as making sure that `totals` (and to some extent `file_totals`) are uptodate.

This now moves all the methods from `EditableReport` onto `Report`, and introduces a new method to iterate over "partially parsed" `ReportFile`s, making sure those are being kept around as well.

---

Part of https://github.com/codecov/engineering-team/issues/3381, and built on top of https://github.com/codecov/shared/pull/548.


tests/benchmarks/test_report.py
```diff
@@ -21,10 +20,9 @@ def read_fixture(name: str) -> bytes:
     pytest.param(Report, False, id="Report"),
     pytest.param(ReadOnlyReport, False, id="ReadOnlyReport"),
     pytest.param(ReadOnlyReport, True, id="Rust ReadOnlyReport"),
-    pytest.param(EditableReport, False, id="EditableReport"),
 ]
 
-EDITABLE_VARIANTS = [Report, EditableReport]
+EDITABLE_VARIANTS = [Report]
```
joseph-sentry, line 25: i think it'd be nice to hardcode this instead of parametrize, since there should only be one variant going forward

joseph-sentry (approved):

PR #550, opened 2025-02-28
fix: Avoid key-error when inserting asset chunks

ajay-sentry:
This PR updates some logic when creating associations to filter out assets that we don't have a mapping for when creating our asset_chunks prior to insertion.

Seeing around 40 instances of the error in the last week so not super wide spread but still worth fixing -> https://cloudlogging.app.goo.gl/Ckj7wYDciPS34nBb6

### Legal Boilerplate

### DIFF
--- a/shared/bundle_analysis/parsers/v3.py
+++ b/shared/bundle_analysis/parsers/v3.py
@@ -188,7 +188,7 @@ class ParserV3(ParserTrait):
                 assert self.session.bundle is not None
                 return self.session.id, self.session.bundle.name
         except Exception as e:
-            # Inject the plugin name to the Exception object so we have visibilitity on which plugin
+            # Inject the plugin name to the Exception object so we have visibility on which plugin
             # is causing the trouble.
             e.bundle_analysis_plugin_name = self.info.get("plugin_name", "unknown")
             raise e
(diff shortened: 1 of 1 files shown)


tests/unit/bundle_analysis/test_asset_association.py
```diff
@@ -50,13 +50,13 @@ def test_asset_association():
...
 
-        # Check that non javscript asset types didn't have their UUIDs updated
+        # Check that non javascript asset types didn't have their UUIDs updated
         for hashed_name, asset in curr_a_asset_mapping_before.items():
             if asset.asset_type != AssetType.JAVASCRIPT:
                 assert asset.uuid == curr_a_asset_mapping_after[hashed_name].uuid
         for hashed_name, asset in curr_b_asset_mapping_before.items():
             if asset.asset_type != AssetType.JAVASCRIPT:
-                assert asset.uuid == curr_b_asset_mapping_after[hashed_name].uuid
+                assert asset.uuid != curr_b_asset_mapping_after[hashed_name].uuid
```
ajay-sentry, line 59: This is so the test fails for now

JerrySentry (approved):

PR #553, opened 2025-03-04
Merge the `Report._files/_chunks`

Swatinem:
Instead of maintaining two separate fields, this merges the previous `_files` which was a dict from filename to `totals`, and the `_chunks` which was a list of split chunks, themselves optionally a fully parsed `ReportFile`.

Maintaining only a single structure for the list of files makes a ton of internals of the `Report` simpler.
It also solves the problem of `file_totals` getting out of sync with the `ReportFile.totals`.

Instead of maintaining a `FileSummary` dataclass, and a separate `ReportFile`, this merges these two structures together, and makes sure that a `Report` always has `ReportFile`s.

Though the new `ReportFile` has a bit more logic around making parsing of its actual `chunk` contents on-demand only when needed, and avoid having to parse things when not strictly necessary.

---

I think the ~40% perf regression is reasonable, as it is doing more work now for a "shallow" parse. The slowness comes from the `ReportFile` constructor being slow. Part of that is because it has to support various arguments and argument types which are primarily only used within tests. In real usage, we either parse `ReportFile`s from json/chunks, or start with an empty file that has lines added/replaced into it.


shared/reports/carryforward.py
```diff
@@ -66,19 +65,16 @@ def generate_carryforward_report(
     Returns:
         EditableReport: A new report with only the info related to `flags` on it, as described above
```
giovanni-guidini, line 66: [very nit] now it returns `Report`, although they might be the same thing?

Docs also mention that it "creates a new report", but the code suggests all changes are done in-place, and the same report is returned.

giovanni-guidini: I wonder if that could pose a problem... say we use `commit.parent` report to build a "new" carryforward report for `commit`. 

Would we save the modified `commit.parent` report back with the changes? Probably not, but if it's possible it might be a good idea to explicit that in the documentation

shared/reports/resources.py
```diff
@@ -285,105 +252,31 @@ def append(self, _file, joined=True):
...
-            if isinstance(lines, ReportFile):
-                return lines
-            report_file = ReportFile(
-                name=filename,
-                totals=_file.file_totals,
-                lines=lines,
-            )
-            if bind:
-                self._chunks[_file.file_index] = report_file
+    def get(self, filename, **kwargs):
```
giovanni-guidini, line 264: It seems that silently ignoring the `bind` argument (and the `_else` to some extent) _could_ lead to unexpected behavior.

I'm sure this way it's easier to integrate with the rest of the code, but maybe it would be better to fail here if people are using those extra options

giovanni-guidini (approved):
Very good change!

shared/reports/resources.py
Swatinem: thats a good point. Python is also a bit annoying with `*args` vs `**kwargs` as well which can cause problems.
regardless of that, I believe the `bind` argument was unused before. I was fixing a bug with that recently as it was completely broken otherwise. So I’m pretty sure noone was using that argument.

I think this PR will need quite a bit of refactoring for downstream code as well, which I haven’t started yet, but will do so soonish.

shared/reports/carryforward.py
Swatinem: Thanks for reminding me about this. Indeed this is a behavior change, and I need to update the docs, and make sure that the calling code is aware of this.

PR #556, opened 2025-03-04
Forward `get_flag_names` for the `ReadOnlyReport`

Swatinem:
The `api_report_service.ReadOnlyReport` overrides the `flags` property, but it looks like the usage of `.flags` would be better served by `get_flag_names` instead.

But that method previously did not exist on that `Report` class, so lets add it.

### DIFF
--- a/shared/reports/readonly.py
+++ b/shared/reports/readonly.py
@@ -103,6 +103,9 @@ class ReadOnlyReport(object):
                 )
         return self._flags
 
+    def get_flag_names(self) -> list[str]:
+        return self.inner_report.get_flag_names()
+
     @property
     def sessions(self):
         return self.inner_report.sessions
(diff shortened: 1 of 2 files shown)


shared/reports/readonly.py
```diff
@@ -103,6 +103,9 @@ def flags(self):
                 )
         return self._flags
 
+    def get_flag_names(self) -> list[str]:
```
joseph-sentry, line 106: should we be caching the result of get_flag_names internally in the `LazyRustReport` or even in the implementation of the inner report in `resource.py`?

Swatinem: I think it might make sense as a followup potentially. Though I guess this is already rather lightweight already, as it it just iterating through the sessions. Though its quite possible that reports have a shitton fo sessions because of perpetually carryforwarding old shit thats never being cleaned up or other kinds of reasons.

PR #560, opened 2025-03-10
Remove a bunch of duplicated tests

Swatinem:
There were "unit" and "integration" tests, which were mostly completely duplicated. With the only exception that the "integration" tests were calling the constructor, and the "unit" test just sets private fields directly.


giovanni-guidini (approved):
I don't understand why these tests are marked `integration`... I know lines are blurred, but what makes them "integration" tests?

The `apply_diff` maybe is calling external services?
Anyway, good to have a slimmer test suite if we don't loose testing power

Swatinem:
I don’t understand either. Add this as another example of why this classification does not make sense.

Also `apply_diff` does not call external services, as you provide the diff directly.

PR #569, opened 2025-03-12
Hardcode the new minio write implementation

Swatinem:
So far, this only removes the feature flag and the branches in code. It does not yet clean up all the now-obsolete code, which will follow later.


giovanni-guidini (approved):
I hope that when we remove the "old" code we can also drop the "new" and "old" names from the functions hehe

joseph-sentry (approved):

PR #570, opened 2025-03-12
Clean up Storage code

Swatinem:
This removes the old storage abstractions including all their tests.

And also inlines the minio adapter read/write implementation back into the base class.

### DIFF
--- a/shared/storage/minio.py
+++ b/shared/storage/minio.py
@@ -7,7 +7,6 @@ from io import BytesIO
 from typing import IO, BinaryIO, Literal, cast, overload
 
 import certifi
-import sentry_sdk
 import urllib3
 import zstandard
 from minio import Minio
@@ -22,7 +21,6 @@ from minio.helpers import ObjectWriteResult
 from urllib3 import HTTPResponse, Retry
 from urllib3.util import Timeout
 
-from shared.metrics import Summary, set_summary
 from shared.storage.base import (
     CHUNK_SIZE,
     PART_SIZE,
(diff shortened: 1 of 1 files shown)


shared/storage/minio.py
```diff
@@ -204,16 +215,58 @@ def write_file(
...
-            is_compressed=is_compressed,
-            compression_type=compression_type,
+            cast(BinaryIO, result),
+            -1,
+            metadata=headers,
+            content_type="text/plain",
+            part_size=PART_SIZE,
+        )
+
+        set_summary(
```
joseph-sentry, line 263: we can probs get rid of these since we're done rolling out, unless we care about monitoring this long term (which i don't think we do tbh)

shared/storage/minio.py
```diff
@@ -16,15 +19,18 @@
...
-from shared.storage.base import BaseStorageService, PresignedURLService
-from shared.storage.compression import zstd_decoded_by_default
-from shared.storage.exceptions import BucketAlreadyExistsError
-from shared.storage.read import new_minio_read
-from shared.storage.write import new_minio_write
+from shared.metrics import Summary, set_summary
+from shared.storage.base import (
+    CHUNK_SIZE,
+    PART_SIZE,
+    BaseStorageService,
```
joseph-sentry, line 29: may be good in a follow up to remove the `BaseStorageService` as well, it was only useful to define an interface when we had multiple implementations, now that we only have one, it's not as usefule, same with the `PresignedURLService` as well

shared/storage/minio.py
```diff
@@ -204,16 +215,58 @@ def write_file(
         is_compressed: bool = False,
         compression_type: str | None = "zstd",
```
joseph-sentry, line 216: i guess this can be done in a follow-up but we should "deprecate" this argument, maybe using overloads and https://typing.python.org/en/latest/spec/directives.html#example

joseph-sentry (approved):
just some comments about possible future improvements

Swatinem:
Thanks for the suggestions. I removed the instrumentation from this PR.

And I agree that merging / removing the abstract base class / inferface is a good followup that I do want to do, along with cleaning up the various usages and parameters. That would likely need a few passes, as it also means having to audit/update all usage sites.

PR #571, opened 2025-03-12
remove FeatureExposure and SimpleMetric models

matt-codecov:
these were never really fleshed out so i am deleting them

if we want to re-add them in the future, we need to do it with partitioning

requires https://github.com/codecov/shared/pull/573 to ship in worker and api before this can merge
also requires https://github.com/codecov/worker/pull/1136 to have shipped in worker, but that's already merged

### Legal Boilerplate


adrian-codecov (approved):
Approved pending tests fixed

Swatinem (approved):

PR #577, opened 2025-03-18
feat: add new ta_timeseries app

joseph-sentry:
we want to be able to set up a separate instance of timescale to separate out coverage timeseries data and test analytics timeseries data


shared/django_apps/ta_timeseries/models.py
```diff
@@ -0,0 +1,104 @@
...
+    failure_message = models.TextField(null=True)
+    framework = models.TextField(null=True)
+    filename = models.TextField(null=True)
+
+    repo_id = models.BigIntegerField(null=True)
+    commit_sha = models.TextField(null=True)
+    branch = models.TextField(null=True)
+
+    flags = ArrayField(models.TextField(), null=True)
+    upload_id = models.BigIntegerField(null=True)
```
Swatinem, line 30: I’m pretty sure I asked this previously, but why are *all* of these (except `outcome`) nullable?

Swatinem (approved):

shared/django_apps/ta_timeseries/models.py
joseph-sentry: i think the only ones that _shouldn't_ be nullable should be `repo_id` `commit_sha` `upload_id` and maybe `computed_name`, all the other ones it's fair game for them to be null

(i think them all being nullable is leftover from my experiments where i wasn't sure what columns were populated when i was moving data from the existing tables to this one)

PR #580, opened 2025-03-20
fix: Nonsentry user on sentry plan should be able to update plan

ajay-sentry:
This PR aims to fix a bug where a user who has not linked their Sentry account tries to update their plan details while they're on the sentry plan. We naively search for the "available plans" for the user adding the sentry plans to that list if they have the account linked. But we should also add the sentry plans to that list if they're currently on a sentry plan

### Legal Boilerplate


calvin-codecov (approved):
LGTM

PR #582, opened 2025-03-21
Hard-enable the `RustReport`

Swatinem:
It turns out that the [`ComparisonProxy.get_impacted_files`](https://github.com/codecov/worker/blob/c87d47b6c64d150f19242c3ef1c3a5821fc0931f/services/comparison/__init__.py#L284-L290) function within `worker` is depending on the `RustReport` being enabled and present.

So lets just remove the feature flag and hard-enable this, to make it clear that other code actually depends on this.

### DIFF
--- a/shared/reports/readonly.py
+++ b/shared/reports/readonly.py
@@ -1,6 +1,4 @@
 import logging
-import os
-import random
 from typing import Any
 
 import orjson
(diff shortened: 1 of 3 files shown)


Swatinem:
Another fun fact:

While the `get_impacted_files` is hardwired to use the `RustReport`, the code to [compare the output of `get_changes` across the python and rust versions](https://github.com/codecov/worker/blob/c87d47b6c64d150f19242c3ef1c3a5821fc0931f/services/comparison/__init__.py#L172-L196) is spewing a ton of log messages about differences between the two 🤦🏻‍♂️

giovanni-guidini (approved):

PR #583, opened 2025-03-21
Remove unused `Profiling*Analyzer` code

Swatinem:
We have deprecated and removed the previous profiling feature and related code/tasks from api and worker a while back. The code here in shared has stuck around until now.


joseph-sentry (approved):
can you open a PR on worker that points to this PR for the shared SHA just to make sure it passes CI

Swatinem:
I don’t think that is strictly necessary, as there is no references in the worker code to this at all.

But sure, I can quickly get one up, why not.

PR #590, opened 2025-03-27
feat: add metadata to MinioStorageService

joseph-sentry:
GCS/S3 objects allow you to add metadata to them when writing, and you can read that metadata back. This is done via request/response headers. The reason I chose to pass in a metadata_container instead of changing the return type of read_file is because I wanted to offer the option to not care about an object's metadata, and this also makes it so this PR won't involve a bunch of changes to the call sites  in worker and API and instead i can just change the call sites i actually care about

### DIFF
--- a/shared/storage/minio.py
+++ b/shared/storage/minio.py
@@ -201,6 +201,7 @@ class MinioStorageService(BaseStorageService, PresignedURLService):
         is_already_gzipped: bool = False,  # deprecated
         is_compressed: bool = False,
         compression_type: str | None = "zstd",
+        metadata: dict[str, str] | None = None,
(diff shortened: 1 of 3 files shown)


shared/storage/minio.py
```diff
@@ -270,6 +291,11 @@ def read_file(
                 )
             raise e
 
+        if metadata_container is not None:
+            for key, value in response.headers.items():
+                if key.startswith("x-amz-meta-"):
+                    metadata_container[key[11:]] = value
```
Swatinem, line 297: is there a `strip_prefix` kind of function? that would be a better idea rather than just slicing an number of chars off the front.

Swatinem (approved):

PR #603, opened 2025-04-11
Optimize `ArchiveField` reads

Swatinem:
This consists of two optimizations:
- It prefers the `archive_field` over the `db_field`, and can thus avoid some queries for the `db_field`, as some queries in API are explicitly `defer`-ing that load.
- It also avoids initializing the `ArchiveService.storage_hash`, which we don’t need for reads (only writes), and which would require us to load the related repository and owner.

### DIFF
--- a/shared/django_apps/utils/model_utils.py
+++ b/shared/django_apps/utils/model_utils.py
@@ -1,6 +1,7 @@
-import json
 import logging
 from typing import Any, Callable, Optional
+
+import orjson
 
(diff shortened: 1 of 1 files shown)


giovanni-guidini (approved):
lgtm... but can we get rid of ArchiveField you think?
It's been quite a long time, do any objects remain that have the value in the DB, not in the archive at this point?

Swatinem:
Probably not in our production environment, but I don’t know about self-hosted. So I am rather not touching this.

